PR #19, opened 2016-09-15
fix links on shoutem.github.io/ui page

ikalafat:
fixes links on https://shoutem.github.io/ui/ page 

screenshot for reference: http://prntscr.com/cic0ix

first link from commit doesn’t lead anywhere, while 2ndlink should lead
to github, but it leads to docs


tenodi:
It's already fixed.

PR #23, opened 2016-09-15
Remove postinstall script

emailnitram:
From the [RNPM](https://github.com/rnpm/rnpm) ReadMe: 

> Now, as you may know, RNPM is merged into React Native core. It means that from now on you don't need to install a third-party software to use your favorite linking functionality (just use a react-native cli).

The postinstall script should no longer be necessary


tenodi:
Did you try building new RN project without this? I think it would break it, because even thought `rnpm` is included into `react-native`, we still need to `link` our fonts and static assets. Maybe consider changing `rnpm link` to `react-native link`.

emailnitram:
You are correct.  I was missing fonts and icons.  Updated new PR to change` rnpm link` to `react-native link`

SoHotSoup:
Sorry, you should create new PR and start from `develop` branch...

PR #26, opened 2016-09-16
Change rnpm link to react-native link

emailnitram:
RNPM is now built in to react native so use react-native link instead of rnpm link in post install script


SoHotSoup:
Great!

SoHotSoup:
💯

PR #28, opened 2016-09-17
Fix defaultSource special case for Android

plougsgaard:
```
<Image source={require('./image.png')} />
```

This relaxes the restriction to allow `Image`s like above.

### DIFF
--- a/components/Image.js
+++ b/components/Image.js
@@ -3,9 +3,13 @@ import {
   Image as RNImage,
   Platform,
 } from 'react-native';
+import _ from 'lodash';
 
 import { connectStyle } from '@shoutem/theme';
 import { connectAnimation } from '@shoutem/animation';
+
+// a valid source is either an object with an uri key or a number (from a `require` call)
+const isValidSource = (source) => _.isNumber(source) || (_.isObject(source) && source.uri)
 
 class Image extends Component {
   static propTypes = {
@@ -23,7 +27,7 @@ class Image extends Component {
 
     // defaultSource is not supported on Android, so we manually
     // reassign the defaultSource prop to source if source is not defined
-    if ((Platform.OS === 'android') && (!source || !source.uri)) {
+    if (Platform.OS === 'android' && !isValidSource(source)) {
       resolvedProps = {
         ...props,
         // Image views are not rendered on Android if there is no image to display,


components/Image.js
```diff
@@ -23,7 +23,7 @@ class Image extends Component {
 
     // defaultSource is not supported on Android, so we manually
     // reassign the defaultSource prop to source if source is not defined
-    if ((Platform.OS === 'android') && (!source || !source.uri)) {
+    if (Platform.OS === 'android' && !source) {
```
zrumenjak: The original condition was here to handle cases when image URL may be undefined. For example, you often have an API response with optional URLs. If we removed the `!source.url` part of the condition then you will have to do something like this in those cases:
```
<Image source={user.avatarUrl ? { uri: user.avatarUrl } : undefined} />
```
instead of simply doing it like:
```
<Image source={{ uri: user.avatarUrl }} />
```

To support both use cases, we should detect if the source is a `require` call, and explicitly handle that case. It looks like the `require` will return a number, based on the prop types from react-native `Image`: https://github.com/facebook/react-native/blob/master/Libraries/Image/Image.android.js#L94-L95.

plougsgaard: I've updated the PR to allow both kinds. 😄

components/Image.js
```diff
@@ -23,16 +23,19 @@ class Image extends Component {
...
-        // Image views are not rendered on Android if there is no image to display,
-        // so we fallback to a transparent image to be compatible with iOS
-        source: defaultSource || require('../assets/images/transparent.png'),
-        // Fixes a bug with local image resizing on Android:
-        // https://github.com/facebook/react-native/issues/4598#issuecomment-162328501
-        style: { ...props.style, width: null },
-      };
+    if (Platform.OS === 'android') {
+      // a valid source is either an object with an uri key or a number (from a `require` call)
+      if (typeof source === 'object' ? !source.uri : typeof source !== 'number') {
```
SoHotSoup: we use lodash as dependency, so I assume you could write this to look more elegant:
```javascript
if (_.isNumber(source) || (_.isObject(source) && source.uri))
```

plougsgaard: Moved your suggestion into a helper function, think it reads more easily now.

PR #76, opened 2016-10-30
Add ability to config onEndReachedThreshold prop

Lughino:


components/ListView.js
```diff
@@ -67,8 +67,10 @@ class ListView extends React.Component {
   static propTypes = {
     autoHideHeader: React.PropTypes.bool,
     style: React.PropTypes.object,
-    data: React.PropTypes.array,
+    data: React.PropTypes.oneOfType([
+      React.PropTypes.object, React.PropTypes.array]),
```
SoHotSoup: When would you use object over array? Could you write an example of passed data as object and what would you expect to be rendered?

Lughino: Certainly, the component [ListViewDataSource](https://facebook.github.io/react-native/docs/listviewdatasource.html#constructor) allows you to use the structure as an array or as an object (map). This means that you can pass a map structured in this way:
```javascript
{
  "postId1": {
     "id": "postId1",
     "title": "lorem ipsum",
     ....
  },
  "postId2": {
     "id": "postId2",
     "title": "lorem ipsum",
     ....
  },
  ...
}
```
Having the data in store with that structure, this means that you should not change them.
Actually it would also be given the option of passing the `getRowData method (dataBlob, sectionID, rowID);` which allows you to instruct the ListView to own structure.

SoHotSoup: Cool! Thank you!

MrBr: If data is object it won't work when grouping data into sections because data is grouped with reduce and object doesn't have reduce [#L41](https://github.com/Lughino/ui/blob/e0873ea5b3d6983a30585cf876ac0c9f4be95be5/components/ListView.js#L41). Update `groupItemsIntoSections` to use lodash [reduce](https://lodash.com/docs/4.16.6#reduce) (it support both object and array).

SoHotSoup:
@MrBr could you take another look at it?

components/ImageGallery.js
```diff
@@ -0,0 +1,356 @@
...
+        style={style.title.container}
+        driver={this.timingDriver}
+        animationName="fadeOut"
+      >
+        <Subtitle style={style.title.text} numberOfLines={2}>{title}</Subtitle>
+      </View>
+    );
+  }
+
+  getImageTransformer(page) {
```
MrBr: Can you keep render methods grouped together, it is much cleaner.

You can also group, update methods, get methods and other.

components/ImageGallery.js
```diff
@@ -0,0 +1,356 @@
...
+
+  renderPage(page, pageId) {
+    const { style } = this.props;
+    const { imageSwitchingEnabled } = this.state;
+    const image = _.get(page, 'source.uri');
+    const title = _.get(page, 'title');
+    const description = _.get(page, 'description');
+
+    if (!image) {
+      return;
```
MrBr: Return `null`? How does this affect gallery? Should we maybe have fallback image?

components/ImageGallery.js
```diff
@@ -0,0 +1,356 @@
...
+    if (event.scale > 1.0 && mode === IMAGE_GALLERY_MODE) {
+      // If controls are visible and image is transformed,
+      // We should switch to image preview mode
+      this.setImagePreviewMode();
+    } else if (mode === IMAGE_PREVIEW_MODE) {
+      this.updateImageSwitchingStatus();
+    }
+  }
+
+  onSingleTapConfirmed() {
```
MrBr: Can you document when will this be called.

components/ImageGallery.js
MrBr: You always pass here `selectedIndex`, better name then for the argument would be `index`.

components/ImageGallery.js
```diff
@@ -0,0 +1,356 @@
...
+    return null;
+  }
+
+  resetImageTransformer(transformer) {
+    transformer.updateTransform({ scale: 1, translateX: 0, translateY: 0 });
+  }
+
+  resetSurroundingImageTransformations() {
+    const { selectedIndex } = this.state;
+    let transformer = this.getImageTransformer(selectedIndex - 1);
```
MrBr: This repeated code should be handled differently.

For example, you can implement `getImageTransformers(indexes)` and then

```
// Surrounding indexes
const indexes = [selectedIndex + 1, selectedIndex - 1]; 
const transformers = this.getImageTransformers(indexes);

transformers.forEach(this.resetImageTransformer);
```

Or you can have `resetImageTransformer` work with `index` and have condition in there (does transformer exists). But I like more the way it is now, it is clean.

components/ImageGallery.js
```diff
@@ -0,0 +1,356 @@
...
+import {
+  ScrollView,
+  LayoutAnimation,
+  InteractionManager,
+} from 'react-native';
+
+import { connectStyle } from '@shoutem/theme';
+import { TimingDriver } from '@shoutem/animation';
+
+import Image from 'react-native-transformable-image';
```
zrumenjak: I wouldn't include it in the toolkit for now. We are currently using this component only in some situations (iOS gallery, and lightbox), but we are also using a different, completely native, component in the gallery on Android.

PR #80, opened 2016-11-01
ability to change color to spinner

Lughino:
With this PR the Spinner component now can set custom color


components/Spinner.js
```diff
@@ -24,6 +26,7 @@ Spinner.propTypes = {
 
 const StyledSpinner = connectStyle('shoutem.ui.Spinner', {
   size: 'small',
+  color: 'gray',
```
SoHotSoup: Could you move this style to `theme.js`

PR #83, opened 2016-11-08
Exponent support

brentvatne:
Hi there! Sorry this took so long :P We just released an update to make react-native-vector-icons play more nicely with Exponent, so now all that's needed is these small changes. We don't support react-native-share either, so maybe at some point that can be removed and replaced with the built-in alternative in React Native, but I didn't bother with that here. Tested it and published to https://getexponent.com/@community/shoutem-ui-examples

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,7 @@
     "react-native-mock": "^0.2.5"
   },
   "scripts": {
-    "test": "mocha -R spec --require test-utils/setup.js --require react-native-mock/mock.js --compilers js:babel-core/register $(find . -name '*.spec.js' ! -ipath '*node_modules*')",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "html-entities": "^1.2.0",
     "htmlparser2": "^3.9.0",
     "lodash": "^4.15.0",
-    "react-native-lightbox": "shoutem/react-native-lightbox",
+    "react-native-lightbox": "git+https://github.com/shoutem/react-native-lightbox#master",
```
brentvatne: this is actually needed for `yarn` support currently

raphaelcosta:
👍

PR #104, opened 2016-12-06
Add inline styleName to NavigationBar

guilhermedecampo:
Fix NavigationBar being hidden by ListView.

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -978,6 +978,13 @@ export default () => ({
       },
     },
 
+  '.inline': {
+    container: {
+      width: window.width,
+        position: 'relative',
+    }
+  },
+
     'shoutem.ui.Title': {
       solidifyAnimation(driver) {
         return {
--- a/theme.js
+++ b/theme.js
@@ -981,7 +981,7 @@ export default () => ({
   '.inline': {
     container: {
       width: window.width,
-        position: 'relative',
+      position: 'relative',
     }
   },
 


theme.js
```diff
@@ -978,6 +978,13 @@ export default () => ({
       },
     },
 
+  '.inline': {
+    container: {
+      width: window.width,
+        position: 'relative',
```
SoHotSoup: Please fix indentation. Remove two extra spaces. Also, have you tested it how does it work in situations like this:
```javascript
render() {
  return (
    <Screen>
      <NavigationBar />
      <ScrollView>
        {/*Some scrollable content*/}
      </ScrollView>
    </Screen>
  );
}
```

I want to know would scroll view's content overlapse navigation bar after scroll... If it would, then add zIndex to navigation bar's style.

Also, could you add this styleName to the documentation as well?

SoHotSoup:
And thank you for submitting the PR!

theme.js
guilhermedecampo: Awesome! Sorry for the indentation =(

Yeah I tested on `ListView`.. going to take a look on `ScrollView` component.

SoHotSoup: Hey, @guilhermedecampo did you find some time to test it? Are we good to go?

guilhermedecampo: Wow so sorry for all this time! 

So I tested with both ListView and ScrollView components.

Check out the gifs below.. 
OBS: Take in consideration that I just changed NavigationBar and that there is a `NavigationHeader` this why the extra space on top.

**Not working - not using inline `styleName`**

navigationbar-

**Working - using inline `styleName` + ListView**

navigationbar-working

**Working - not using inline `styleName` + ScrollView**

navigationbar-working-scrollview

SoHotSoup:
@guilhermedecampo Thank you man!

Issue #132, opened 2017-01-17
Restaurant exemple error


Slumber86:
Checked-out the example project and launching 
```
yarn
react-native link
react-native run-ios
```
I'm reaceving this error message:
![simulator screen shot 17 gen 2017 17 06 37](https://cloud.githubusercontent.com/assets/7034226/22031970/c92ccc4a-dce2-11e6-88e2-f7eebe3686d2.png)
text in debugger console:

```
Warning: Failed prop type: The prop `colors` is marked as required in `LinearGradient`, but its value is `undefined`.
    in LinearGradient (created by LinearGradient)
    in LinearGradient (created by AnimatedComponent)
    in AnimatedComponent (created by Styled(AnimatedComponent))
    in Styled(AnimatedComponent) (created by NavigationBarView)
    in RCTView (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent (created by NavigationBarView)
    in NavigationBarView (created by _class)
    in _class (created by AnimatedComponent)
    in AnimatedComponent (created by Styled(AnimatedComponent))
    in Styled(AnimatedComponent) (created by NavigationTransitioner)
    in ScrollDriverProvider (created by NavigationTransitioner)
    in RCTView (created by View)
    in View (created by NavigationTransitioner)
    in RCTView (created by View)
    in View (created by NavigationTransitioner)
    in RCTView (created by View)
    in View (created by NavigationTransitioner)
    in NavigationTransitioner (created by RNCardStack)
    in RNCardStack (created by CardStack)
    in CardStack (created by Styled(CardStack))
    in Styled(CardStack) (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by Root)
    in Provider (created by Root)
    in Root
    in RCTView (created by View)
    in View (created by AppContainer)
    in RCTView (created by View)
    in View (created by AppContainer)
    in AppContainer
Cannot read property 'map' of undefined
```
my environment:

```❯ node --version
v6.9.1

~
❯ yarn --version
0.17.10

~
❯ react-native --version
react-native-cli: 2.0.1
react-native: 0.39.0

~
❯
```

PR #133, opened 2017-01-18
Remove default navbar gradient style

MrBr:


MrBr:
https://github.com/shoutem/ui/issues/132

Issue #142, opened 2017-01-25
Release failed because of react-native-share 1.0.18


g3r4n:
XCode Release failedb because of new header introduce with RN 0.40.0
React native : 0.39.2

rsekulic1:
@g3r4n Thank you for submitting the issue and sorry for a late response. We're currently working on fixing the majority of the issues and we will release a new version of `@shoutem/ui` hopefully next week with the fix included.

SoHotSoup:
Fixed with latest release.

PR #143, opened 2017-01-25
Update package.json

g3r4n:
make it work with RN 0.39.2 in Release scheme in XCode


g3r4n:
correct the issue #142

g3r4n:
better way to avoid issue is to upgrade to latest version of RN

PR #153, opened 2017-02-07
Add custom icon to dropdown button

pvpshoot:
Hi, this is my first PR and i want to help make this world better =)

In this PR i make an opportunity to customise icon in DropDownMenu.

U can choose between icon name or pass the component to render it.

Please tell me if i do something wrong.

### DIFF
--- a/components/DropDownMenu.js
+++ b/components/DropDownMenu.js
@@ -44,6 +44,13 @@ class DropDownMenu extends Component {
      */
     valueProperty: React.PropTypes.string.isRequired,
     /**
+     * Icon name or component to render on opening button
+     */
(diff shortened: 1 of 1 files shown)


components/DropDownMenu.js
```diff
@@ -137,13 +144,14 @@ class DropDownMenu extends Component {
 
   renderSelectedOption() {
     const { style, titleProperty } = this.props;
+    const getIcon = (icon = drop-down) => (_.isString(icon) ? <Icon name={icon} /> : <icon />)
```
zrumenjak: Define this function outside of a render function. This will unnecessarily create a new function on each render.

Issue #165, opened 2017-02-14
HorizontalPager and TabBar


martinezguillaume:
I think that we need to provide a way to display a tab bar depending of the HorizontalPager like this: 

![demo-fb](https://cloud.githubusercontent.com/assets/17292331/22953453/93435e06-f311-11e6-88c4-93a13eb76a1b.gif)

For that, the props "renderOverlay" of HorizontalPager needs to provide a function in parameter that go to the index that we specified. (```goToPage(index)``` for example).

the props "renderOverlay" need to provide too the values of the function 'OnScroll' of the ScrollView to have the liberty to add animation in the tab bar depending of the scroll values.

We can have a look in this library https://github.com/skv-headless/react-native-scrollable-tab-view

Please share if you have others ways to do it !

SoHotSoup:
@martinezguillaume I have noticed that you are very active here, thank you for that, I will take a look at possible solutions and write it here. Also, I will include @ikalafat, who originally wrote the component into this discussion.

martinezguillaume:
Hello ! I forked the repo here
https://github.com/martinezguillaume/ui/tree/feature/horizontalpager-tabbar
I'm working on it 😁

SoHotSoup:
Cool! I will write down my thoughts on it here later.

martinezguillaume:
I've made a pr here https://github.com/shoutem/ui/pull/170, I know there is a lot of enhancement but I think that it's more efficient to work from a base 😁

ikalafat:
@SoHotSoup I've checked the PR (not-so-thorougly, yet) and wrote my comments/thoughts there, but your wisdom and thoughts will be appreciated, too :)

Vladimir-Vdovic:
Checked PR #170, assuming this is now resolved, so I will close this thread. Prompt for re-opening if issue is still relevant.

PR #170, opened 2017-02-17
Enhance HorizontalPager with a TabBar

martinezguillaume:
ref: https://github.com/shoutem/ui/issues/165

This pr has to purpose to give access to you to had Tab bar with your Horizontal Pager like that:
demo-fb

For that, the prop renderOverlay provide more informations about the ScrollView:

- scrollToPage: A function to scroll to a given index ```scrollToPage(index)```
- scrollValue: An Animated.Value (from 'react-native) that has the scroll offset
- selectedIndex: A number that is the index where you are right now

The thing is that you have to listen to the ScrollValue to update your render depending of the offset.

To listen to the scrollValue: (I'm using recompose)
```js
  withState('scroll', 'setScroll', 0),
  lifecycle({
    componentDidMount() {
      this.props.scrollValue.addListener(scrollValue =>
        this.props.setScroll(scrollValue),
      );
    },
  }),
```

Now, you have the offset of the scrollView inside the state ```scroll.value``` and it'll update your render each time that scrollValue change !

Now, use this value to display correctly your component of the tabbar, maybe you need to calculate some more stuff if you want to modify the color, or the opacity, or the position of the tabs.

For example, I'll give you an example if you want to change the opacity. The selected tab will have an opacity of 1, the others of 0.30.
Here is my function:
```js
function tabOpacity(scroll, activeTab, currentTab) {
  const disabledOpacity = 0.30;

  const page = floor(scroll);
  const offset = ((scroll - page) * (1 - disabledOpacity)) + disabledOpacity;

  if (currentTab === activeTab) {
    if (currentTab === page) {
      return (1 + disabledOpacity) - offset;
    }
    return offset;
  }

  if (currentTab === activeTab - 1) {
    if (currentTab === page) {
      return (1 + disabledOpacity) - offset;
    }
    return disabledOpacity;
  }

  if (currentTab === activeTab + 1) {
    if (currentTab === page + 1) {
      return offset;
    }
    return disabledOpacity;
  }

  return disabledOpacity;
}
```

Then for each tab, I'll use this function to display the correct opacity and I'll assign an onPress function that redirect you in the correct index
```js
      map(TAB_HOME_MENU, (name, i) =>
        $(View, {
          styleName: 'flexible',
          style: {
            opacity: tabOpacity(scroll.value, selectedIndex, i),
          },
        },
          $(Button, { onPress: () => scrollToPage(i) },
            $(Icon, { name }),
          ),
        ),
      ),
```

And here we go 🎉 (it's a little laggy because I'm in developement mode)
fevr -17-2017 20-30-46

You can disable or not the animation during the scrollToPage with the ```animated``` prop of HorizontalPager. If it's false, i'll just redirect without animation to the specified index, if it's true, it'll make the scrolling animation !

I know that we can enhance it more ! But it's just to start on something 😊 
(I've made a little fix in ListView)

The library https://github.com/skv-headless/react-native-scrollable-tab-view was a very good help !


ikalafat:
Hey @martinezguillaume 

Thanks for activity! Much appreciated.

I haven't checked over the whole diff yet, but I'll quote few things to get things moving.

`animated` prop in `defaultProps` should default to `false` so that it doesn't break some our internal components :), however I like the idea. 

Regarding this:
```
For that, the prop renderOverlay provide more informations about the ScrollView:

scrollToPage: A function to scroll to a given index scrollToPage(index)
scrollValue: An Animated.Value (from 'react-native) that has the scroll offset
selectedIndex: A number that is the index where you are right now

```

I don't see a point in modifying renderOverlay (at least not at this moment, but it is late here :) ) in this manner (plus it will break some of our internal stuff). Overlay should be "overlay" and it should fill its parent. Its main purpose (in our usecase) is to allow users to render `PageIndicators` (dots) that usually represent number of `pages` in pager (or in general, to allow users to render something over pages). I think that you shouldn't use Overlay to render tabbar, it should be separated (explanation below) 

If I understand you correctly, you need a way to programmatically change `selectedIndex` from outside, right? If that is the case, I would rather modify `componentWillReceiveProps` and add something like this

```
if (this.props.selectedIndex !== nextProps.selectedIndex) {
  this.scrollToPage(nextProps.selectedIndex);
}
```

That way you wouldn't need to expose scrollToPage method, right? 

As for `scrollValue`, I'm currently thinking that it would be better to expose additional prop `onScroll` which would be called at the end of `onHorizontalScroll` method, and which would forward the `native event` from `onHorizontalScroll(event)`

```
onHorizontalScroll(event) {
   const { onScroll } = this.props;
   ...a lot of existing code
   if (_.isFunction(onScroll)) {
     onScroll(event) // maybe other arguments like width / pageMargin  - need to think.
   }
}
```

My suggestion would result in something like this in your code (simplified, just that you get my idea):

```
<View>
  <TabbarContainer>
    <Tab onPress={()=>{ this.setState({ selectedIndex: 0 })}}>
    ...content..
    </Tab>
    <Tab onPress={()=>{ this.setState({ selectedIndex: 1 })}}>
    ...content..
    </Tab>
  </TabbarContainer>
  <HorizontalPager
    data={...}
    animated
    selectedIndex={this.state.selectedIndex}
    onScroll={...here you would take care of your tabbar animation stuff...}
    onIndexSelected={(selectedIndex)=>{ this.setState({ selectedIndex })}}
  />
</View>
```

If you tap on one of tabs, horizontal pager would go to page related to your tab. If you swipe between pages, you would still have mechanism to detect swipe gesture, do some math and update state on your screen (plus renderOverlay would be intact) 

Let me know what you think - i'll try to go deeper into this early next week, but feel free to ask any questions you might have.

martinezguillaume:
I really like this approach !
I agree with you with the scrollToPage called automatically when selectedIndex is changed in the componentWillReceiveProps ! It'll be so much simplier to not worried about that, just give the selectedIndex you want and the component will do it for you !

I was a little worried about manipulate the scrollValue inside our whole component because it means that each time it changes, the tabbar and the scrollView will re-render, but the scrollView will re-render no matter what right ? So there is no problem to put it here, I'll test and I think that I'll be more able to see more enhancements ! Thanks ! ☺

ikalafat:
Glad to hear that you like suggestion! :)

As for manipulating the `scrollValue` in component, I'm not sure will it cause re-render on each update, however you can try it.

Easiest way to check it is to place `console.log("rendering");` in `render` method and to check how many times will you get "rendering" string in console :)

If you're seeing a lot of those strings, maybe we can optimize it a bit - i'll think a bit about this.

martinezguillaume:
I have few problems

For now, the component is used like that:

HorizontalPager
```js
  withState('selectedIndex', 'setSelectedIndex', 0),
  withState('scrollValue', 'setScrollValue', new Animated.Value(0)),
...
  return $(Screen, null,
    $(View, { styleName: 'flexible' },
      $(HorizontalPager, {
        ...
        scrollValue,
        selectedIndex,
    ),
    $(HomeTabBar, {
      scrollValue,
      selectedIndex,
      setSelectedIndex,
    }),
  );
```

TabBar:
```js
  withState('scroll', 'setScroll', 0),
  lifecycle({
    componentDidMount() {
      this.props.scrollValue.addListener(scrollValue =>
        this.props.setScroll(scrollValue),
      );
    },
  }),
```

So, I have to initialize directly the Animated.Value in my component and it'll be set by HorizontalPager in each onScroll
```js
  onHorizontalScroll(event) {
    ...
    // We're setting the value of the prop scrollValue
    if (!_.isNil(scrollValue)) {
      scrollValue.setValue(contentOffset.x / width);
    }
  }
```

I explain,
If we use an onScroll func called with scrollValue in parameter, we need to stock this value in your state to use it in your TabBar component (because now, they are separated). We have our first issue here because on each ```onHorizontalScroll```, it'll set your state causing high lagging. That's we're using an Animated.Value object. It's more optimize !

So, despite this, we have another issue.
The use of state ```selectedIndex``` inside our component to manage the scrollToPage is very very good ! But this state doesn't change if we change index in the HorizontalPager without calling ```setSelectedIndex``` in your TabBar.

So, I see two options on my side.
We stay with the props ```renderOverlay``` that provide you the props needed (```selectedIndex```, ```scrollToPage```, ```scrollValue```) and we try to not brake anything (with the pageIndicators for example). But we loose the logic with your ```selectedIndex```that detects automatically if the HorizontalPager need to scroll.

We add the prop ```onScroll``` called in ```onHorizontalScroll``` and provide you the ```selectedIndex```. So, you'll have thing like that
```js
onScroll: selectedIndex => setSelectedIndex(selectedIndex),
```

If you have other way to do it, don't hesitate 😃 sorry for the long post 😔

martinezguillaume:
So ? 😃

martinezguillaume:
I've merge it, the renderOverlay provide two more arguments, the ```scrollToPage``` function and
the ```scrollValue```

prototype:
```js
renderOverlay(data, selectedIndex, scrollToPage, scrollValue)
```

SoHotSoup:
Hey @martinezguillaume!

I have finally come to this, and I have the same problem, as you.

But I'm more into this solution:

#### 1st step: Remove `selectedIndex` from components state
If selected index would always be controlled from the outside any component could scroll to page.

Changing this [line](https://github.com/shoutem/ui/blob/develop/components/HorizontalPager/HorizontalPager.js#L104) read from `nextProps` instead from `this.state` would be enough.

#### 2nd step: introduce `onScroll` as @ikalafat already mentioned. That way we could attach any animated value to original scroll event and achieve animated page indicators, for example a line which is sliding from page to page as you scroll.

Let me know if you need any more clarification or don't have time to implement it. I need that soon, so if you are packed, I can do it.

Thanks for being patient!

martinezguillaume:
I can do it if you want ! But with this, we can't manage good the redirection into a page when you press the related icon.

For example, if you click on the 'News' tab, it'll scroll to this page
demo-fb

It looks like your solution manage good the animation, but not the redirection into the good page when you click a tab

SoHotSoup:
I have tried it locally:
demo

This is just without `onScroll` Also, it would be great if we could use `ViewPager` as base component on android and `ScrollView` on iOS. I can do it, but I feel uncomfortable with taking your contribution. It is really great how much you are trying to help, so I'm here to help you as much as you need with it.

martinezguillaume:
No no no no don't worry ! You're not taking anything ! Just do it because I don't really understand what you've did, don't worry 😊

SoHotSoup:
I will fork your solution, and extend it to preserve your contributions... and then we will merge it from my fork, so your hard work will pay off.

yusufyildirim:
Thank you for your passion guys. I was decided to do that but as i saw you've already done that. I'll be waiting for merge. You'll see my name on PRs soon, too ^^. Thank you again for your efforts and this beautiful repo!

JulianKingman:
What happened to this PR, it looks like it didn't get merged?

JulianKingman:
On a related note, shouldn't I be able to access methods like "scrollToIndex" via the HorizontalPager's ref? (It doesn't seem to work).

This is what I'm trying:
```javascript
goFirstPage() {
  // error: scrollToIndex undefined
  this.pager.scrollToIndex(1)
}
render() {
  return <HorizontalPager ref={c => this.pager = c} />
}
```

martinezguillaume:
I don't know what happened to this pr 😂for your problem, try to use scrollToPage function

PR #202, opened 2017-03-21
Hook a brotha up

GentryRiggen:
Allow me to easily change just theme variables.

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -141,7 +141,14 @@ const defaultVariables = {
   sectionHeaderBackgroundColor: '#F2F2F2',
 };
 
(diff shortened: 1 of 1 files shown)


SoHotSoup:
please pull the upstream, this is already done within 0.11.0

zrumenjak:
I have exported the `defaultThemeVariables` in #219, this should allow you to change only some variables by doing something like this in your app:
```
import { getTheme, defaultThemeVariables } from '@shoutem/ui';

const theme = getTheme({
  ...defaultThemeVariables,
  featuredColor: 'green',
});
```

Issue #203, opened 2017-03-22
What's happening with NavigationBar component?


yourcontainer:
Clean installation
```javascript
import React, { Component } from 'react';
import { AppRegistry } from 'react-native';
import { NavigationBar, Icon, Title } from '@shoutem/ui';

export default class app extends Component {
  render() {
    return (
        <NavigationBar
            leftComponent={<Icon name="sidebar" />}
            centerComponent={<Title>TITLE</Title>}
        />
    );
  }
}

AppRegistry.registerComponent('app', () => app);
```
```
"dependencies": {
    "@shoutem/ui": "^0.12.0",
    "react": "15.4.2",
    "react-native": "0.42.3"
  },
```

Result
![3f120e1fbb](https://cloud.githubusercontent.com/assets/13626192/24176626/c2ff097e-0ea5-11e7-81ce-24b7ab34c9f3.jpg)

yusufyildirim:
Hi,

What is the exact problem? Can you explain? So people can help you...

kheide:
The text is out of wack compared to what the screenshots in the tutorial. Seems to be a CSS issue.

filiphosko:
The centerComponent (in this case 'TITLE') overrides it's vertical alignment set by the parent, but the other components (leftComponent, rightComponent) don't. You can check it [here](https://github.com/shoutem/ui/blob/develop/theme.js#L1292). The parent sets 'flex-end' to all of the components, but centerComponent has 'alignSelf' set to 'center'. I think centerComponent should just comply with what the parent set on it, the question is if it is better to leave 'flex-end' or 'center' set for every component (I think 'center' looks better).

zrumenjak:
Navigation bar issues are now fixed in develop. The fixes will be published to npm in the next version.

PR #211, opened 2017-03-30
Remove 'alignSelf' from centerComponent

filiphosko:
Since centerComponent overrides 'flex-end' vertical alignment set by it's parent and sets it to 'center' for itself, it breaks vertical alignment with it's siblings (leftComponent, righComponent). centerComponent should just receive vertical alignment by it's parent and not override it.

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -1289,7 +1289,6 @@ export default (variables = defaultVariables) => ({
     },
 
     centerComponent: {
-      alignSelf: 'center',
       alignItems: 'center',
       flex: 1,
     },


filiphosko:
This PR adresses this issue https://github.com/shoutem/ui/issues/203 .

zrumenjak:
Thanks for the PR, but this didn't solve the problem completely. I have tested your fix, and it still caused alignment issues when the navigation bar had action buttons. I have fixed this issue a bit differently in #216.

filiphosko:
It's true that I didn't test it with Button element, only with TouchableOpacity, it worked that way. Great you solved it!

PR #217, opened 2017-04-05
Fix incorrect button styles and outdated examples

zrumenjak:


aswinmohanme:
How do I use this fix right now?, or should I wait for another release ?

zrumenjak:
This fix will be included in the next release. I expect that we will release a new version tomorrow.

If you need this fix right away, a workaround is to call `getTheme` and provide custom theme variables to that function. You can then apply the colors that I have changed in this pull request. After that, you will also have to use a `secondary` styleName in situations where you previously used `dark`. You can see examples of that in the `Examples` component.

zrumenjak:
@aswinmohanme this fix has not been released in v0.13.0: https://github.com/shoutem/ui/releases/tag/v0.13.0

ng-hai:
Should update docs about Button `styleName`

Issue #150, opened 2017-02-02
Listview OnRefresh spins forever, how do I complete the refresh status


mypark:
How do I set the refresh status to complete so that the spinner stops and the pull to refresh bounces back?

I implemented the onRefresh function on the ListView property and return a result but it still keeps spinning.  I don't see anywhere to finish the refresh and reset the status.

      ```
     <ListView
          data = { results }
          autoHideHeader = { true }
          loading = { isLoading }
          onRefresh = { async () => {
            console.log('refreshing');
            const results = await refetch();
           // we have results data but we are still spinning ,how to finish spinning?
          // fresh data has already been rendered at this point but refresh control still visible
          } }
          renderHeader = { this.renderHeader }
          renderRow= { this.renderRow }
          renderSectionHeader = { this.renderSectionHeader }
          getSectionId = { this.getSectionHeaderId }
        />
    ```

note that isLoading is a redux connected property set by the apollo client which seems to work fine for the loading spinner, just not the refresh spinner.  If I set loading to a state variable instead - it seems to work.  Why is that?

brandon-dewitt:
@mypark I just experienced this issue also.  It looks like the ListView state stats goes into 'refreshing', but has no mechanism to come out of refreshing:

https://github.com/shoutem/ui/blob/develop/components/ListView.js#L131

crpr:
I have the same problem. Any workaround?

zrumenjak:
You should use the `loading` prop of the `ListView`. Setting the `loading` prop to `false` will dismiss the spinner.

It could be nice if we also had an option that `onRefresh` can return a promise, and that the spinner is automatically dismissed when that promise is resolved or rejected. PRs are welcome, if anyone wants to implement that.

buncismamen:
anyone have example code for this?, 

@zrumenjak how to set the loading props?, isn't props should be immutable?

zrumenjak:
@buncismamen yes, the idea is to change the `loading` prop of the `ListView` based on some external value. Your example should work, but you are now manually controlling the loading status. In a real world use case, you would probably set the `loading` state value to `false` when a network request is finished.

azadious:
Must set to this.setState({loading: true}) before changing to this.setState({loading: false}).

If you just want to try how it works, use this code.

`onRefresh() {
        console.log('refresh');
        this.setState({loading: true}, function () {
            this.setState({restaurants: restaurants, loading: false});
        });

    }
`

PR #232, opened 2017-04-17
Use Promises  for RefreshControl

bluetec:
I faced with an issue that in `ListView` _RefreshControl_ spins forever...
Someone recommend to use ListView loading prop to Dismiss refreshControl, but it seems in ListView Statuses are strings like `Status.IDLE`, `Status.REFRESHING` and so on, but the loading prop is a boolean value and it cause issues.

Currently refresh Control automatically set _ListView Component_ `status: Status.REFRESHING` and then call `this.props.onRefresh` that I was passed from MyComponent.
In MyComponent I have an state like `this.state.loading : false` and when I want to set `loading` prop, I should set it to `this.setState({loading: true})` on MyComponent `onRefresh` prop, so as soon as ListView `componentWillReceiveProps` triggered, refreshControl will disappear ( in Less than one second ), _because my status state is not equals to ListView component status state_. So there is no possibility to keep spins without hacks.

so Its nice to use Promise to handle this issue.
```
<ListView
    data={this.state.items}
    loading={this.state.loading}
    onRefresh={() => {
       return new Promise((resolve, reject) => {
           setTimeout(() => {
               resolve(true);
           }, 4000);
       });
    }}
    renderRow={() => {}}
/>
```

### DIFF
--- a/components/ListView.js
+++ b/components/ListView.js
@@ -132,7 +132,16 @@ class ListView extends React.Component {
     });
 
     if (this.props.onRefresh) {
-      this.props.onRefresh();
+      const callback = this.props.onRefresh();
+      const setIdle = () => {
+        this.setState({
+          status: Status.IDLE,
+        });
+      };
+
+      if(typeof callback === 'object' && typeof callback.then === 'function') {
+        callback.then(setIdle, setIdle);
+      }
     }
   }
 


bluetec:
#150

components/ListView.js
```diff
@@ -132,7 +135,18 @@ class ListView extends React.Component {
     });
 
     if (this.props.onRefresh) {
-      this.props.onRefresh();
+      const callback = this.props.onRefresh();
+      const setIdle = () => {
+        if(this._isMounted) {
```
SoHotSoup: Do we really need this? Is it even possible that `onRefresh` is called before component is mounted?

bluetec: It because there is no way to _unsubscribe_ from promises, so `this._isMounted` check make sure that setState will not execute when component is UnMounted.

buncismamen:
any update on this?

PR #240, opened 2017-04-25
-

sstimac:


sstimac:
@MrBr After taking another look, it seems that the _.pick works correctly. We only had style in props, from the list of possible PropTypes, so that's what we got. Still, it's weird they don't have the pageSize added in....

MrBr:
Add missing keys to the keys array, and document why are they added manually.

MrBr:
@sstimac can we resolve this?

Issue #241, opened 2017-05-01
undefined is not an object (evaluating '_reactNative.NavigationExperimental.Header')


aswinmohanme:
Error:
`undefined is not an object (evaluating '_reactNative.NavigationExperimental.Header') ` gets thrown when the app is built and run.

Steps I Took:
     Initialize an empty react native project
    Add `yarn add @shoutem/ui'
   `react-native link`
    Use the '<Examples />` Component
    `react-native run-android`

My System Configuration
```
react-native --version                                                                                                                      22:10:05
    react-native-cli: 2.0.1
    react-native: 0.44.0
```

Package.json
```
{
	"name": "testShoutem",
	"version": "0.0.1",
	"private": true,
	"scripts": {
		"start": "node node_modules/react-native/local-cli/cli.js start",
		"test": "jest"
	},
	"dependencies": {
		"@shoutem/ui": "^0.13.0",
		"react": "16.0.0-alpha.6",
		"react-native": "0.44.0"
	},
	"devDependencies": {
		"babel-jest": "19.0.0",
		"babel-preset-react-native": "1.9.1",
		"jest": "19.0.2",
		"react-test-renderer": "16.0.0-alpha.6"
	},
	"jest": {
		"preset": "react-native"
	}
}

```

michaelspeed:
I am still getting this error! any fix?

michaelspeed:
I saw ur commit now there is a new error!!! can't find variables!! theme.js:621:34

dobrienlzd:
Try this for now:  

edit package.json  -> Replace 0.44.0 with 0.43.4   for react-native

'rm -rf node_modules'  //get rid of previous modules 
npm install
react-native link
react-native run-android    // but in  separate window be running   'npm start' first.

aswinmohanme:
So this is how I resolved it. 

Copy the Entire File Contents from the attached commit to the theme.js file in node_modules/@shoutem/ui. 

Replace `variabls` with `variables`

JulianKingman:
Oops, sorry for the typo, I fixed it.

nori3tsu:
NavigationExperimental has been removed from React Native 0.44.0, so I think Shoutem UI should remove and replace it.

https://github.com/facebook/react-native/pull/13066

aswinmohanme:
For the Time Being 
` yarn add shoutem/ui#242/head `

JulianKingman:
[...]

PR #242, opened 2017-05-01
Removed NavigationExperimental dependency

JulianKingman:
Fixes this issue: https://github.com/shoutem/ui/issues/241

RN 0.44.0 eliminated NavigationExperimental, so I copied their style for navigation bar height and put it into the default variables to make it overwritable.


SoHotSoup:
Thank you for submitting it, but this issue would persist for anyone using `CardStack` and `NavigationBar` from `@shoutem/ui/navigation`. The real solution would be to migrate them to use `react-navigation`, but I believe that this is too much to do for now. So I suggest to take the `NavigationExperimental` files from RN 0.42.0 and copy them here. It's little bit hacky, but it would fix all navigation issues on RN 0.44.0. What you say?

SoHotSoup:
#244

JulianKingman:
According to FB the views exist in react-navigation
````
NavigationExperimental is deprecated and will be removed in a future 
version of React Native. The NavigationExperimental views live on in 
the React-Navigation project, which also makes it easy to declare
navigation logic for your app. Learn more at https://reactnavigation.org/
````

**Comment by JulianKingman** [comment]
I replaced navigationexperimental with react-navigation, is there a repo that I can test it with? I don't have any projects that use the header or cardstack.
Still todo: draw in the header height from the Header component like it was originally.

**Comment by SoHotSoup** [comment]
@JulianKingman RestaurantsApp in examples uses it, so please try to run an app and verify that everything works...

**Comment by JulianKingman** [comment]
Hm, this might take some work. They changed the API for the cardstack in the most recent version, I'm thinking of just switching to the new API. If I can get away with just fixing the current errors, I will (tracking here: https://github.com/react-community/react-navigation/issues/1446)

**Comment by SoHotSoup** [comment]
@JulianKingman Sorry, I was offline few days. Thank you for doing this issue, is there anything I can do to help you?

**Comment by davidvuong** [comment]
Hey guys, what's the status on this? I'm currently referencing the latest commit in this PR until this is merged. I'm under the impression that some components won't work even with the fix now? Is it possible for me to get a list of working components?

**Comment by JulianKingman** [comment]
The problem are the navigation components, which rely on NavigationExperimental, which was removed from react in 0.44.0 (I think).
The crux of the issue is that the components were ported to `react-navigation`, but now they have a different API, so I'm not sure if it will be possible to make it fully backward compatible. @SoHotSoup what are your thoughts, is backwards compatibility a must? I think the main change is using `react-navigation`'s built in navigation props, instead of passing in a redux state specifically.

**Comment by jcguarinpenaranda** [comment]
Please merge this!! I am using react-navigation and I am getting the error: undefined is not an object (evaluating '_reactNative.NavigationExperimental.Header')

**Comment by alxvallejo** [comment]
+1, I rely on this PR, but it is causing issues when I use a <Tile> component:

> ExceptionsManager.js:71 Warning: Failed prop type: Invalid props.style key `heroAnimation` supplied to `AnimatedComponent`.

**Comment by tafelito** [comment]
+1, same as @alxvallejo I get the same warning

**Comment by SeptiyanAndika** [comment]
+1, same as @jcguarinpenaranda , i get that error. (evaluating '_reactNative.NavigationExperimental.Header')

**Comment by JulianKingman** [comment]
Waiting on @SoHotSoup to decide what to do about backwards-compatibility of navigation components.

**Comment by AWDGroup** [comment]
Any update on this?

**Comment by Talor-A** [comment]
to anyone having this issue try deleting `~/.npm/` and your `node_modules/` folder. this was causing the wrong version of react-native to install for me.

**Comment by juji** [comment]
+1

**Comment by elopez89** [comment]
Any updates on this?

**Comment by JulianKingman** [comment]
I'm looking into what's involved in using the updated API, assuming that works for @SoHotSoup. Unfortunately, there is no documentation yet (tracking here: https://github.com/react-community/react-navigation/issues/1946), so it's kind of trial and error..

**Comment by JulianKingman** [comment]
@SoHotSoup after looking more in depth into the various API's and changes, it seems like the most sensible solution will be to remove navigation from shoutem/ui. I updated my branch accordingly, as well as the restaurant app. The result is a significant simplification of the code.

For posterity, here are all the potential options I can think of:
1. Create a polyfill for the way react-navigation deals with `CardStack` to match current shoutem/ui API. This would be fragile and likely to break in a new release.
2. Copy relevant code from `NavigationExperimental` into shoutem/ui core. This seems like a dead end, but at least it would be stable.
3. Take navigation out of shoutem/ui. This seems preferrable to me, as navigation is quite complex (even in the Shoutem implementation), react-navigation is the future, and it's relatively simple to use.

**Comment by eightyfive** [comment]
My humble opinion is to go for option 3. I don't think shoutem UI Toolkit should include any navigation logic, just UI. Pure UI framework. But this sounds unrealistic.

4. A more viable solution is to migrate to [react-navigation](https://github.com/react-community/react-navigation)

@SoHotSoup Any plan to do so? Need help? What's the plan at Shoutem to deal with this issue? I guess this is actively being looked into, since you whole ecosystem cannot upgrade to new RN updates.

**Comment by AWDGroup** [comment]
I agree with @eightyfive and @JulianKingman, Shoutem UI should only contain the UI components and not the navigation or any other logics.

**Comment by AkdM** [comment]
While I totally agree about having only the UI and not the logic, would this mean that components like the DropdownMenu would get kicked out since there’s some logic in it?

**Comment by eightyfive** [comment]
@AkdM I don't think Shoutem will ever have a plan to ditch the navigation logic from the Toolkit. It would impact too much their existing product.

Let's wait for an input from someone working at Shoutem.

**Comment by AkdM** [comment]
@eightyfive My guess would be to have two different repos or even branches, one with the fixed Navigation and the other one without.

**Comment by eightyfive** [comment]
Yes, I would like that as well 😄. But unfortunately this is not in our hands.

**Comment by SoHotSoup** [comment]
@eightyfive we have opted for 3rd solution and we are actively working on moving our main product to use react-nativation. I guess that would be finished and deployed in next few weeks, as part of that task we will remove these navigation components from our UI.

**Comment by AkdM** [comment]
Great news! Thanks for the update @SoHotSoup 👍

**Comment by pavitran** [comment]
@SoHotSoup any update on this?

**Comment by kiodou** [comment]
any update? i get an error just by importing

**Comment by leandrodragani** [comment]
what happen with this ?

**Comment by davidvuong** [comment]
Judging by this commit: https://github.com/shoutem/ui/commit/3bc085622430d9929e3d97dddad3446cb8b8f603

shoutem/ui now might be compatible with later versions of React Native which no longer have `NavigationExperimental`? I haven't had a chance to test it myself.

**Comment by Vladimir-Vdovic** [comment]
We've resolved it with the commit davidvuong mentioned. I wanted to update the thread immediately but I didn't get a chance to test it out completely.

I can now run the `<Examples />` without any of the previous workarounds (dropping RN version, manually importing NavigationExperimental, etc.).

Here's my `package.json` dependency versions:
```
{
  "name": "UITest",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "jest"
  },
  "dependencies": {
    "@shoutem/ui": "^0.21.0",
    "react": "16.0.0-alpha.12",
    "react-native": "0.47.1"
  },
  "devDependencies": {
    "babel-jest": "20.0.3",
    "babel-preset-react-native": "2.1.0",
    "jest": "20.0.4",
    "react-test-renderer": "16.0.0-alpha.12"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

**Comment by JulianKingman** [comment]
PR no longer necessary

**Comment by princenaman** [comment]
On a fresh react-native init
Still getting:
```js
warning "@shoutem/ui@0.21.1" has incorrect peer dependency "react@^15.0.0".
warning "@shoutem/theme@0.10.0" has incorrect peer dependency "react@^15.1.0".
warning "@shoutem/animation@0.11.1" has incorrect peer dependency "react@^15.0.0".
```

package.json:
```js
{
  "name": "sample",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "start": "node node_modules/react-native/local-cli/cli.js start",
    "test": "jest"
  },
  "dependencies": {
    "@shoutem/ui": "^0.21.0",
    "react": "16.0.0-alpha.12",
    "react-native": "0.47.1"
  },
  "devDependencies": {
    "babel-jest": "20.0.3",
    "babel-preset-react-native": "2.1.0",
    "jest": "20.0.4",
    "react-test-renderer": "16.0.0-alpha.12"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

bhosle-git:
Hi,

i'm facing the same issue - does the fix work with RN 0.44 also or do i have to upgrade to RN 0.47?

thanks

PR #263, opened 2017-05-31
Add missing padding to NavigationBar container

juanreyesv:
[This changes](https://github.com/shoutem/ui/pull/216/files) fixed the vertical alignment but removed the padding so the icons on the Navbar didn't have any spacing to the sides.


Vladimir-Vdovic:
Looks good to me!

Will merge it into v0.23.3 hotfix!

PR #269, opened 2017-06-15
Feature/seext 4508 navbar image

dgrubelic:
Refactored navigation bar children composers to a mode dynamic architecture.

### DIFF
--- a/navigation/NavigationBarView.js
+++ b/navigation/NavigationBarView.js
@@ -341,10 +341,6 @@ class NavigationBarView extends PureComponent {
   }
 
   createNavigationHeaderProps(style) {
-    if (this.props.navigationBarImage || NavigationBarView.navigationBarImage) {
-      this.props.backgroundColor = 'transparent';
-    }
-
     const headerProps = {
       ...this.props,
       style: [navigationHeaderStyle, style.navigationHeader],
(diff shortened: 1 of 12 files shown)


theme.js
```diff
@@ -1464,6 +1468,25 @@ export default (variables = defaultThemeVariables) => ({
       right: 0,
       height: NAVIGATION_BAR_HEIGHT,
     },
+    navigationBarImage: {
+      flex: 1,
+      flexGrow: 1,
+      height: NavigationHeader.HEIGHT,
+      left: 0,
+      solidifyOpacityAnimation(driver) {
```
SoHotSoup: why `solidifyOpacity`? This would be applied only if someone gives `animationName="solidifyOpacity"` to navBarImage.

navigation/NavigationBarView.js
```diff
@@ -368,6 +368,15 @@ class NavigationBarView extends PureComponent {
     return null;
   }
 
+  renderBackgroundImage() {
```
SoHotSoup: My idea was to have `renderBackground` prop and background image composer would compose `renderBackground` that renders `Image`. That way we would be able to remove `fade` styleName and pass `renderBackground` which renders gradient. That makes us closer to ability of removing react-native-linear-gradient from ui toolkit. cc @zrumenjak

Issue #255, opened 2017-05-13
TextInput: focus is not a function


seblau:
I do have a couple of `TextInput` fields for user registration and I would like to implement a `next` button on the keyboard, so that the user can focus the next `TextInput` field from the previous one. I did it with `refs` like described here: https://facebook.github.io/react-native/docs/textinput.html

When I use the `TextInput` component from `shoutem/ui`, I do get an error message (see screenshot) that the function `focus` is not defined. However when I import the `TextInput` component from `react-native` it works.

```
import { TextInput } from '@shoutem/ui';

_focusNextField(nextField) {
  this.refs[nextField].focus()
}

render() {
  <TextInput
            ref='1'
            placeholder="Email"
            returnKeyType="next"
            blurOnSubmit={false}
            onSubmitEditing={() => this._focusNextField('2')}
  />
  <TextInput
            ref='2'
            placeholder="First name"
            returnKeyType="next"
            blurOnSubmit={false}
  />
}
```

![img_6033](https://cloud.githubusercontent.com/assets/3884582/26025403/0026df7e-37e7-11e7-9574-d32914eefc95.png)

Vladimir-Vdovic:
May be related to:
https://github.com/shoutem/ui/issues/44

Vladimir-Vdovic:
A short update.

This improvement is being worked on via [this pull request](https://github.com/shoutem/ui/pull/280), which aims to expose React Natives text input.

Issue #44, opened 2016-09-26
TextInput ref methods are broken


patrick-michelberger:
I try to call the focus() method on a TextInput ref. Such as

```javascript 
<TextInput ref="emailInput" onSubmitEditing={() => this.focusNextField('passwordInput')}/>
<TextInput ref="passwordInput"/>

// component method
focusNextField = (nextField) => {
  this.refs[nextField].focus();
};
```

However I get the following Error Message:
`_this.refs[nextField].focus is not a function`

The Reference Type seems to be different compared to the React Native core and doesn't offer methods such as `focus()`

dculjak:
Hi Partric, this obviously is a bug/missing functionality. We're putting it in our roadmap for one of the next releases.

lijy91:
Is there any temporary solution?

JulianKingman:
Yes, if you copy the TextInput from this library (https://github.com/shoutem/ui/blob/develop/components/TextInput.js), add a ref at line 22, like so:
````javascript
<RNTextInput
  {...props}
  style={style}
  placeholderTextColor={props.style.placeholderTextColor}
  selectionColor={props.style.selectionColor}
  ref={(input) => this.props.inputRef(input)}
/>
````
Then, using that textinput, you can use the `inputRef` prop as you would a normal ref, like this:
````javascript
<TextInput inputRef={(input) => { this.textInput = input }} />
````
The input ref is now available via `this.textInput`

neegool:
I'm trying out the temporary solution above, and I can't get it to work. I get the following error message: `_this2.props.inputRef is not a function`.

neegool:
Okay, I found out that by making my own copy of TextInput , and adding a `ref="textInput"` prop to it, I was able to access the TextInput via `this.refs[nextField].wrappedInstance.wrappedInstance.refs.textInput`. This will work for me, for now. :)

JulianKingman:
It sounds like you used a string based ref, its recommended to use a callback ref instead.

steven-tib:
Hi, any progress on this bug? 
I'm trying to set focus on my next textInput but cannot make it work. 
thanks

jenskuhrjorgensen:
For anyone ending up here struggling with focussing the next TextInput:
https://stackoverflow.com/a/45253129/3749432

itanhduy:
This issue is still there. Kindly help us. Because this function is very basic.

danalloway:
fixed in #471

[...]

PR #280, opened 2017-07-02
Add inputRef prop

Jkaveri:
Exposes the react native's TextInput.

#255 #44

### DIFF
--- a/components/TextInput.js
+++ b/components/TextInput.js
@@ -6,9 +6,9 @@ import { connectAnimation } from '@shoutem/animation';
 
 class TextInput extends Component {
   render() {
-    const { props } = this;
+    const { inputRef, ...props } = this.props;
     const style = {
-      ...props.style,
+      ...props.style
     };
     delete style.placeholderTextColor;
     delete style.selectionColor;
@@ -18,6 +18,7 @@ class TextInput extends Component {
       <RNTextInput
         {...props}
         style={style}
+        ref={inputRef}
         placeholderTextColor={props.style.placeholderTextColor}
         selectionColor={props.style.selectionColor}
         underlineColorAndroid={props.style.underlineColorAndroid}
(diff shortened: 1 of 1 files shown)


JulianKingman:
Can this be merged? This is a super important PR for things like navigating between fields, and focusing/blurring fields programmatically.

zrumenjak:
@Jkaveri thanks for the PR, but we have plans to address this issue in a different way. I would like to expose all the methods from the underlying `TextInput` component on the connected component, so that it could be used in the same way that the original `TextInput` is used, i.e., by just capturing the ref to it and calling method on that object.

We have started testing this approach in a limited number of cases, you can see examples in https://github.com/shoutem/ui/blob/develop/components/ScrollView/ScrollView.js#L88. Ideally we would like to implement a helper function (`hoistMethods`) that would work similarly to `hoistStatics`, but just for methods.

This function would:
* go through all the methods of the underlying component
* create the methods with the same name on the target component that call the original methods on the ref to the underlying component

With this method implemented, we could solve all the related issues for ScrollView, ListView, and any other React Native component that has instance methods.

I will close this PR, it would great if you could implement something like this, if you have time. Feel free to contact me if you have any questions.

JulianKingman:
Will this work? https://github.com/elado/hoist-non-react-methods

zrumenjak:
@JulianKingman we need something similar to the package that you linked. That package has two issues that I can see:
1. It hoists static methods, and throws an error if the method already exists
  `connect` from redux (and many other HoC implementations) already do this using `hoistStatics`, 
  and this results in errors if you start using this package when there are components in the 
  hierarchy with already hoisted statics.

2. It trows an error when a method already exists on the target class
  It would be better to simply skip those methods, because it was probably implemented on the 
  target class  to override the underlying behaviour.

`hoist-non-react-methods` would probably work, if it ignored statics, and methods that already exist.

JulianKingman:
I've made a fork that solves 1, I'll add a check for 2 and try to get it merged.

JulianKingman:
I finally got around to making the PR, you can test with my fork: https://github.com/DesignmanIO/hoist-non-react-methods, hopefully it gets merged soon.

JulianKingman:
@zrumenjak The upstream repo has been updated, but I tried just 'tossing it in', and couldn't quite figure it out. Can you take a look?
https://github.com/elado/hoist-non-react-methods/

dodsky:
@JulianKingman I'll take a look and see if we can leverage that package, it seems that both concerns have been fixed:
 1: https://github.com/elado/hoist-non-react-methods/blob/master/src/index.js#L52 hoist statics is now optional
2: already defined methods won't get hoisted https://github.com/elado/hoist-non-react-methods/blob/master/src/index.js#L78

Issue #321, opened 2017-10-16
Could not add StyleSheet to <TextInput>


JoshYuJump:
When I add StyleSheet to <TextInput>, it failed.
```
<TextInput
    placeholder={'Password'}
    style={styles.inputText}
/>
```
`inputText` is defined in the following code.
```
const styles = StyleSheet.create({
  inputText: {
    flex: 1, 
    height: 30, 
    lineHeight: 30, 
    paddingVertical: 0
  },
})
```

It success if I write like that:
```
<TextInput
    placeholder={'Mobile'}
    style={{flexGrow: 1, height: 30, lineHeight: 30, paddingVertical: 0}}
/>
```

thallysrc:
I reproduced this bug , and the `const styles = StyleSheet.create()` creates a styles.inputText with a number type instead of object

if you insert this console 
` console.log(typeof styles.inputText)` 
after this
```
const styles = StyleSheet.create({
  inputText: {
    flex: 1, 
    height: 30, 
    lineHeight: 30, 
    paddingVertical: 0
  },
})
```
you'll see the type of styles.inputText are `Number` and the component TextInput expect `object` you can remove the StyleSheet.create()
```
    const styles = {
      inputText: {
        flex: 1, 
        height: 30, 
        lineHeight: 30, 
        paddingVertical: 0,
        backgroundColor: 'red'
      },
    }
```
it works for me, but i'll make a pull request to fix this. peace!

Vladimir-Vdovic:
Closing issue, you can find the solution/workaround [here](https://github.com/shoutem/ui/pull/322#issuecomment-352032215).

PR #322, opened 2017-10-17
Fixed issue #321

JoshYuJump:


JoshYuJump:
Any update or advise @dodsky

components/HorizontalPager/HorizontalPager.js
```diff
@@ -58,6 +58,10 @@ class HorizontalPager extends Component {
     scrollEnabled: PropTypes.bool,
     // Style prop used to override default (theme) styling
     style: PropTypes.object,
```
dodsky: You can't have two style definitions. Remove the one on L60.

components/Video/Video.js
```diff
@@ -50,6 +50,10 @@ class Video extends React.Component {
       uri: PropTypes.string,
     }),
     style: PropTypes.object,
```
dodsky: Same comment as above.

JoshYuJump: Which type of `style` is right as your opinion, object or number?
If I create a style like use `StyleSheet.create`, It will return a number.

dodsky: Your prop types are OK, but you didn't remove the old style declaration, so you now have this:
```
style: PropTypes.object, // remove this
style: PropTypes.oneOfType([
  PropTypes.object,
  PropTypes.number,
]),
```
Just remove the first one, please.

JoshYuJump:
@dodsky , Hi, I have fixed the duplicate definitions.

dodsky:
Thanks. Merging.

Vladimir-Vdovic:
@JoshYuJump 

After looking into it and testing it out, `@shoutem/theme` and `@shoutem/ui` isn't compatible with `StyleSheet.create`.

A poor workaround would be to utilize `StyleSheet.flatten`, which uses the js<->native bridge twice as opposed to `StyleSheet.create`, making it slower.

A better alternative is to create a JavaScript style and pass it as the `style` prop instead of using `StyleSheet.create`.

Of course, you can always make your own custom styles using `@shoutem/theme`, one good example can be found in [this Stack Overflow answer](https://stackoverflow.com/a/46058338/7920643).

JoshYuJump:
Well, I see! You can reject my PR. :)

PR #384, opened 2018-03-12
rendering ListView when some of the render functions change

### LINKED ISSUE #383: ListView doesn't update when any of the render functions change
IMO the ListView should update when any of the render functions change.

The right place to do it is to validate the render functions on the lifecycle method `shouldComponentUpdate(nextProps, nextState)`.

rcacheira:
fixes #383


ingoclaro:
In this case why not extend from PureComponent instead?

rcacheira:
That is a good suggestion, I just followed the approach defined later. I'll check and propose the change if it will not conflict with the rest of the ListView functionality.

PR #477, opened 2020-04-29
remove expo, add locks and bump other deps

sstimac:
initial dependency clear PR


scripts/add-native-deps.js
```diff
@@ -1,29 +1,33 @@
...
+        ...nativeDeps,
+      }
+    };
 
-  fs.writeJsonSync(ROOT_PACKAGE_JSON_PATH, newPackageJson, { spaces: 2 });
-  console.log('[@shoutem/ui] - native dependencies added to root package.json');
+    fs.writeJsonSync(ROOT_PACKAGE_JSON_PATH, newPackageJson, { spaces: 2 });
+    console.log('[@shoutem/ui] - native dependencies added to root package.json');
+  } catch (e) {
+    console.log('[@shoutem/ui] - skipping adding dependencies to root package.json');
```
Definitely-Not-Vlad, line 28: Could we add actual error logging here instead of just skipping? Or is the error useless in cases of this that you've encountered?

sstimac: Roger that, will display ze error

PR #505, opened 2020-07-14
optimize animating and other logic

sstimac:
- Removed animation wrappers from animation package, that were used for dropdown
- optimized rerenders on both ListView and DropDown modal ( we did rerender on every option layout change and similar )
- deferred to using local Modal animation instead of Animated API ( Modal has issues with using native driver, and animating without this doesn't really make sense
- optimized dropdown modal render strategy to center short content via padding. Flexbox centering in this case wouldn't work as Android doesn't respect flex values properly in case of flatlist and it's content wrapper
- optimize gradient values and calculations
- remove archaic shouldComponentUpdate on FlatList to allow rerender on style changes and similar. Removed with built in PureComponent

Above should result in a couple of fixes as well as significat android performance improvement

### DIFF
--- a/components/DropDownMenu/DropDownModal.js
+++ b/components/DropDownMenu/DropDownModal.js
@@ -2,12 +2,9 @@ import PropTypes from 'prop-types';
 import React, { PureComponent } from 'react';
 import {
   Modal,
-  LayoutAnimation,
   Dimensions,
 } from 'react-native';
 import _ from 'lodash';
-
-import { TimingDriver, FadeIn, ZoomOut } from '@shoutem/animation';
 import { connectStyle, changeColorAlpha } from '@shoutem/theme';
 
 import { Button } from '../Button';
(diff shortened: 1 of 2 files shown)


sstimac:
@toroz This is what i mean in terms of updating UI tooklit. This is something we can find of pretty much every component we're using. Shoutem Animation could also use an overhaul

components/DropDownMenu/DropDownModal.js
```diff
@@ -232,16 +214,16 @@ class DropDownModal extends PureComponent {
...
+      location: Math.max((bufferHeight / screenHeight), 0.2),
       color: invertedColor
     }, {
-      location: (bufferHeight + gradientHeight + transparencyHeight) / screenHeight,
-      color: invertedColor,
+      location: Math.min((bufferHeight + listViewHeight) / screenHeight, 0.8),
+      color: invertedColor
     }, {
-      location: Math.max((bufferHeight + listViewHeight) / screenHeight, 0.75),
+      location: Math.min((bufferHeight + listViewHeight + gradientHeight) / screenHeight, 0.85),
```
Definitely-Not-Vlad, line 226: Could we extract these decimal values into constants with descriptive names? Just so it's possible to understand what exactly these values are supp.osed to be and why they are such values. Ideally you could add a comment for each resulting `const`

PR #524, opened 2020-09-28
Feature/seext 8410 rss photos details screen title cut off

tomislav-arambasic:
Added padding to ImageGalleryOverview's title for iOS devices.
Android already had this title under the image counter. Did the same for iOS devices.
Now, if the title is 2 rows long, it won't go over the image counter.

Note: image counter is barely visible because it gains color from the navigation title color setting, so it's the same color as the ImageGallery component background.

Before:
image

After:
IMG_0061
IMG_1406
IMG_1407

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.0.4",
+  "version": "4.0.5",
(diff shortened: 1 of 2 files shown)


majaklajic (approved):
Looks OK. Don't merge it until you publish new UI version...

PR #528, opened 2020-10-27
Feature/android fontfamily resolving

Definitely-Not-Vlad:
`resolveFontFamily` takes the `fontFamily`, `fontWeight` and `fontStyle` properties provided via Customize style settings page in the Shoutem Builder, or from variables being provided in any other way.

Also overriding `fontStyle` and `fontWeight` for Android because having these properties set to anything other than `normal`, while using Italic, Bold or BoldItalic fonts will result to the app using the default system font.

Screenshot showing successful loading of fonts on Android with a mix of `normal`, `bold` and `italic` with `NotoSarif`.

Screenshot 2020-10-27 at 23 50 54

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -52,8 +52,35 @@ export function dimensionRelativeToIphone(dimension, actualRefVal = window.width
   return getSizeRelativeToReference(dimension, 375, actualRefVal);
 }
 
-// This function is deprecated and replaced with calculateLineHeight. 
-// It remains present here because of the backward compatibility. 
+// 'fontWeight' and 'fontStyle' aren't always supplied for every component, so we're setting default
+// values of 'normal'.
+export function resolveFontFamily(fontName, fontWeight = 'normal', fontStyle = 'normal') {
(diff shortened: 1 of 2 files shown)


theme.js
```diff
@@ -376,20 +406,41 @@ export default (variables = defaultThemeVariables) => ({
 
     lineHeight: formatLineHeight(variables.heading.fontSize),
     ...variables.heading,
+    fontFamily: resolveFontFamily(
+      variables.heading.fontFamily,
+      variables.heading.fontWeight,
+      variables.heading.fontStyle,
+    ),
+    fontWeight: Platform.OS === 'ios' ? variables.heading.fontWeight : 'normal',
```
majaklajic, line 414: Would it make sense to extract this into the separate function since it's being used accross the theme?

majaklajic (approved):

theme.js
Definitely-Not-Vlad: Yes it would, done and done. Great call!

PR #537, opened 2020-11-05
Add error message and border styling to TextInput

Definitely-Not-Vlad:
- added `errorText` default variables to `theme.js`
- added `highlightWhenFocused` boolean prop which adds `borderWidth: 1` to `TextInput` when it is focused
- added `errorMessage` string prop which renders a `Caption` with `errorText` styling under the given `TextInput`

Focused input highlighting:
Screenshot 2020-11-05 at 15 14 59

Error `Caption` with `errorText` styling (not default, being overridden by Shoutem builder settings):
Screenshot 2020-11-05 at 15 15 11

### DIFF
--- a/components/TextInput.js
+++ b/components/TextInput.js
@@ -32,11 +32,15 @@ class TextInput extends PureComponent {
     const { highlightWhenFocused, errorMessage, style } = this.props;
     const { isFocused } = this.state;
 
-    const resolvedStyle = {
+    let resolvedStyle = {
       ...style,
-      borderWidth: (isFocused || errorMessage) ? 1 : 0,
-      borderColor: style.errorBorderColor,
-    };
+      borderWidth: (isFocused || !!errorMessage) ? 1 : 0,
+    }
+
+    if (!!errorMessage) {
+      resolvedStyle.borderColor = style.errorBorderColor;
+    }
+
     delete resolvedStyle.placeholderTextColor;
     delete resolvedStyle.selectionColor;
     delete resolvedStyle.underlineColorAndroid;
--- a/components/TextInput.js
+++ b/components/TextInput.js
@@ -10,6 +10,16 @@ import { connectStyle } from '@shoutem/theme';
 import { Caption } from './Text';
 import { View } from './View';
 
+// Style properties defined in theme which would case a warning if passed to a components style prop
+const omittedStyleProperties = [
+  'errorBorder',
+  'placeholderTextColor',
+  'selectionColor',
+  'underlineColorAndroid',
+  'withBorder',
+  'withoutBorder',
+];
+
 class TextInput extends PureComponent {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


components/TextInput.js
```diff
@@ -1,35 +1,88 @@
...
-    const { props } = this;
-    const style = {
-      ...props.style,
+  constructor(props) {
+    super(props);
+
+    autoBindReact(this);
+
+    this.state = {
+      isFocused: props.highlightWhenFocused && props.autoFocus,
```
sstimac, line 19: Careful with autoFocus prop here. If there's multiple text inputs with autoFocus, only one of them will be focused. Maybe make use of something like this to detect if the field is actually focused -> https://reactnative.dev/docs/textinput#isfocused .

sstimac: Also, let's go for a shorter prop name, something like highlightOnFocus

components/TextInput.js
```diff
@@ -1,35 +1,88 @@
...
+      resolvedStyle.borderColor = style.errorBorderColor;
+    }
+
+    delete resolvedStyle.placeholderTextColor;
+    delete resolvedStyle.selectionColor;
+    delete resolvedStyle.underlineColorAndroid;
+
+    if (highlightWhenFocused) {
+      return {
+        onFocus: () => this.handleFocus(),
```
sstimac, line 50: no need for arrow fns here. Just use shorthand syntax -> onFocus: this.handleFocus,

components/TextInput.js
```diff
@@ -1,35 +1,88 @@
...
+  handleBlur() {
+    this.setState({ isFocused: false });
+  }
+
+  resolveProps() {
+    const { highlightWhenFocused, errorMessage, style } = this.props;
+    const { isFocused } = this.state;
+
+    let resolvedStyle = {
+      ...style,
```
sstimac, line 36: you can use https://lodash.com/docs/4.17.15#omit here to define base style. That way you don't have to use `delete` later on in the function.

components/TextInput.js
sstimac: Come to think of it, do we really need to conditionally pass the focus and blur handlers here ? Have the handlers do this logic, and just always pass the handlers, which will decide to react or not.

generally we don't like the dynamic prop calculation that happens on every render. Sometimes we use it for style, but i would limit this concept to style calculation only. Even then, i would prefer style resolution inside render, like this

style={[style.baseStyle, !!errorMessage && style.errorBorder, isFocused && style.focusedBorder]} ... It's easier to read, and it would use existent styles conditionally. Provided that they are created with StyleSheet, it would bring an additional benefit of optimization, whereas we're creating a new style object on every render here.

sstimac (changes requested):
This really looks like something now :) Good job !

I have only a few comments in terms of syntax

components/TextInput.js
Definitely-Not-Vlad: Much like a regular React Native `TextInput`, it's up to the developer to make sure there's only one `TextInput` with `autofocus`.

Will change the name.

components/TextInput.js
Definitely-Not-Vlad: I sincerely disagree that it's easier to read, it also doesn't take into account `highlightOnFocus` prop.

If I'm understanding your comment about optimization correctly, it doesn't apply to our use case or how we work.

components/TextInput.js
Definitely-Not-Vlad: `isFocused()` method requires refs which does not seem at all worth the implementation in my humble opinion.

sstimac: What's the problem with refs ? It shouldn't be that hard to implement. It would provide you with the exact and true information, that is immune to developer mistake. I believe that will only lead to a better code.

components/TextInput.js
sstimac: it was just an example. You should fortify the condition based on your case. 

The fact that we're not passing our styles through StyleSheet doesn't mean we won't in the future. The fact that we're avoiding doing things properly because of the provided reason only makes it harder to implement the improvement in the future, as we keep piling on the code that will be harder to change.

sstimac (approved):
:*

components/TextInput.js
```diff
@@ -1,28 +1,69 @@
...
-      <RNTextInput
-        {...props}
-        style={style}
-        placeholderTextColor={props.style.placeholderTextColor}
-        selectionColor={props.style.selectionColor}
-        underlineColorAndroid={props.style.underlineColorAndroid}
-      />
+      <View>
+        <RNTextInput
+          {..._.omit(this.props, 'style')}
```
toroz, line 49: this is usually done like this: `const { errorMessage, highlightOnFocus, style, ...otherProps } = this.props;` and then here you'd just say `{...otherProps}`.
I think this would be the correct approach here as we don't want to send `errorMessage` nor `highlightOnFocus` to the RNTextInput anyway.

components/TextInput.js
```diff
@@ -1,28 +1,69 @@
...
+          onBlur={this.handleBlur}
+          onFocus={this.handleFocus}
+          placeholderTextColor={style.placeholderTextColor}
+          selectionColor={style.selectionColor}
+          underlineColorAndroid={style.underlineColorAndroid}
+          style={[
+            _.omit(style, omittedStyleProperties),
```
toroz, line 56: Again, I'd opt here for the approach: `const { placeholderTextColor, selectionColor, ..., ...stylesheetStyle } = style;` 
And then you'd say `placholderTextColor={placeholderTextColor}` (and so on for the rest of them) and `style={...stylesheetStyle, ...}`

I'm not a fan of this `omittedStyleProperties` array because we're gonna have to maintain it.

PR #538, opened 2020-11-09
Feature/empty list component

majaklajic:
Added `EmptyListView` component that is similar to `EmptyStateView`, but with more attractive illustration. Here are a few screenshots of how it looks. Also, I'm open for your suggestions and copy updates. 

android
emptyListView_prime
emptyListView_rose
emptyListView_noir
emptyListView_bleu

+ I've also updated ListView component, introduced possability to add custom `contentContainerStyle` style to our existing style from theme.

### DIFF
--- a/components/EmptyListView.js
+++ b/components/EmptyListView.js
@@ -31,12 +31,12 @@ class EmptyListView extends PureComponent {
   }
 
   render() {
-    const { message, title, imageStyle, titleStyle, messageStyle } = this.props
+    const { message, title, imageStyle, titleStyle, messageStyle, style } = this.props
     const EmptyStateImage = this.resolveImage();
 
     return (
-      <View styleName="vertical h-center">
-        <EmptyStateImage styleName='image' style={imageStyle} />
+      <View styleName="vertical h-center ">
+        <EmptyStateImage style={imageStyle || style.image} />
         <Title styleName="title" style={titleStyle}>{title}</Title>
         <Text styleName="description" style={messageStyle}>{message}</Text>
       </View >
--- a/components/EmptyListView.js
+++ b/components/EmptyListView.js
@@ -1,0 +1,30 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connectStyle } from '@shoutem/theme';
+import { Text, View } from '@shoutem/ui';
+import EmptyList from '../assets/images/emptyList.svg'
+
+function EmptyListView({ message, title }) {
+
+  return (
+    <View styleName="vertical h-center">
+      <EmptyList style={style.image} />
+      <Text style={style.title}>{title}</Text>
+      <Text style={style.description} styleName="h-center">
+        {message}
+      </Text>
+    </View >
+  );
+}
+
+EmptyListView.propTypes = {
+  ...EmptyListView.propTypes,
+  message: PropTypes.string,
+  title: PropTypes.string,
+};
+
+const StyledView = connectStyle('shoutem.ui.EmptyListView')(EmptyListView);
+
+export {
+  StyledView as EmptyListView,
+};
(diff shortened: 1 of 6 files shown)


components/EmptyListView.js
```diff
@@ -0,0 +1,38 @@
...
+    title: "It's empty in here",
+    message: "We couldn't find anything to show...",
+  }
+
+  render() {
+    const { message, title, style } = this.props
+
+    return (
+      <View styleName="vertical h-center">
+        <EmptyList style={style.image} />
```
sstimac, line 18: I'd prefer dynamic values for this. I may not want chat type of image for every list

components/EmptyListView.js
```diff
@@ -0,0 +1,38 @@
+import PropTypes from 'prop-types';
```
sstimac, line 1: let's use this new component inside FlatList directly. -> https://reactnative.dev/docs/flatlist#listemptycomponent No need to conditionally render anymore

components/EmptyListView.js
sstimac: Also, it would be cool to support animation here. more of a design task, but maybe Poriluk can hook us up

sstimac, line 18: Let's use another name for this graphic. having list in it's name indicates it is a more complex list component

sstimac (changes requested):
noice. A couple of suggestions

components/EmptyListView.js
majaklajic: Agree

components/EmptyListView.js
majaklajic: Great idea... That's exactly where I was going to use this.

components/EmptyListView.js
majaklajic: Any suggestions? `EmptyStateView` is already taken, so maybe `ClearStateView`, `ZeroDataStateView`...

sstimac: EmptyListImage ? using image keyword would provide a lot of context, as we're using it as a named component

components/EmptyListView.js
```diff
@@ -0,0 +1,61 @@
...
+    message: "We couldn't find anything to show...",
+  }
+
+  constructor(props) {
+    super(props);
+
+    autoBindReact(this);
+  }
+
+  resolveImage() {
```
sstimac, line 21: nice. I was going to suggest the usage of defaultProps, but this is even better

components/EmptyListView.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  render() {
+    const { message, title, imageStyle, titleStyle, messageStyle } = this.props
+    const EmptyStateImage = this.resolveImage();
+
+    return (
+      <View styleName="vertical h-center">
+        <EmptyStateImage styleName='image' style={imageStyle} />
+        <Title styleName="title" style={titleStyle}>{title}</Title>
```
sstimac, line 40: Just a suggestion. You don't need to implement if you don't find it better. I wouldn't go for stylenames here as it's actually a default style. I'd go for something like this <Title style={[style.title, titleStyle]}>{title}</Title>

For me, styleName would be something like "bold" or "italic" in this example, meaning a variation that stands out from default style. So for our components, you would simply override the base style if you need to, instead of editing a variation that you may need later on

theme.js
```diff
@@ -1544,6 +1544,30 @@ export default (variables = defaultThemeVariables) => ({
       };
     },
   },
+  'shoutem.ui.EmptyListView': {
+    'shoutem.ui.View': {
```
sstimac, line 1553: if you decide to go for suggested approach from above, this style will be much simpler, and won't depend on component composition, meaning you could reuse it in multiple locations, not just in this specific component sequence

sstimac (approved):
noice !!

components/EmptyListView.js
majaklajic: I've decided to go with your suggestion. It looks simple and it allows toolkit users to tweak style much easier. Thank you!

PR #542, opened 2020-11-12
Use spread operator for leftover props and style

Definitely-Not-Vlad:
Use spread operator for leftover props and style instead of using `_.omit`.

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -1967,9 +1967,15 @@ export default (variables = defaultThemeVariables) => ({
     fontWeight: resolveFontWeight(variables.text.fontWeight),
     fontStyle: resolveFontStyle(variables.text.fontStyle),
 
-    errorBorderColor: variables.errorText.color,
-    withBorder: 1,
-    withoutBorder: 0,
+    errorBorderColor: {
+      borderColor: variables.errorText.color,
+    },
+    withBorder: {
+      borderWidth: 1,
+    },
+    withoutBorder: {
+      borderWidth: 0,
+    },
   },
 
   'shoutem.ui.NumberInput': {
--- a/theme.js
+++ b/theme.js
@@ -1967,16 +1967,9 @@ export default (variables = defaultThemeVariables) => ({
     fontWeight: resolveFontWeight(variables.text.fontWeight),
     fontStyle: resolveFontStyle(variables.text.fontStyle),
 
-    errorBorder: {
-      borderWidth: 1,
-      borderColor: variables.errorText.color,
-    },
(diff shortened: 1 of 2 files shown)


theme.js
```diff
@@ -1967,16 +1967,9 @@ export default (variables = defaultThemeVariables) => ({
...
-    },
-    withBorder: {
-      borderWidth: 1,
-    },
-    withoutBorder: {
-      borderWidth: 0,
-    },
+    errorBorderColor: variables.errorText.color,
+    withBorder: 1,
+    withoutBorder: 0,
```
toroz, line 1972: I would suggest we revert this to the way it was before. 
We should aim to have "plain CSS" as much as possible. Plain CSS would mean we use CSS props (borderWidth) instead of introducing contextual knowledge that "withBorder: 1" actually means that borderWidth css property will be 1.

toroz (changes requested):

theme.js
Definitely-Not-Vlad: I agree and tried that first, however, the only way I managed to make this work was by using this. The reason I have to make a separate object is because I can't have a style in theme with two different `borderWidth` defined.

Two alternatives are:
1. Give the separate object a CSS property:
```
withBorder: {
  borderWidth: 1,
},
```

And then applying it as `borderWidth: withBorder.borderWidth`.
Currently:
```
style={{
  ...otherStyle,
  borderWidth: hasBorder ? style.withBorder : style.withoutBorder,
  borderColor: !!errorMessage && style.errorBorderColor,
}}
```
After changes:
```
style={{
  ...otherStyle,
  borderWidth: hasBorder ? withBorder.borderWidth : withoutBorder.borderWidth,
  borderColor: !!errorMessage && errorBorderColor.borderColor,
}}
```

2. Create style names for the `TextInput` component and then resolve the style apply said `styleNames`.

I would opt for the 1st option as I'm currently unsure whether the 2nd option would actually lend itself to the `StyleSheet` optimization we talked.

toroz: How about
```

style={{
  ...otherStyle,
  ...(hasBorder ? style.withBorder : style.withoutBorder),
}}
```

Definitely-Not-Vlad: I'm assuming that's going with the 1st option where this is in place:
```
withBorder: {
  borderWidth: 1,
},
```

If so, then yes, I don't see why it wouldn't work. I'll try when I get some time and commit the changes if it works.

toroz (approved):

PR #545, opened 2020-11-23
Hide ListEmptyComponent while loading and fix ListEmptyComponent in s…

majaklajic:
- Hid `ListEmptyComponent` while loading `ListView` data
- Added different empty state handling for `SectionList` 

`SectionList` displays `ListEmptyComponent` only when it receives an empty array, not empty data => https://github.com/facebook/react-native/issues/14721#issuecomment-311771858


Definitely-Not-Vlad (approved):
LGTM

PR #550, opened 2020-12-04
Update TextInput.js

StephenTANM:
Fix for onFocus & onBlur silently failing. Reproduction example below...

`  
import { TextInput } from "@shoutem/ui";

const [state, setstate] = useState<InitialState>({
    focused: false,
  });

  const onFocus = () => {
    setstate((oldstate) => ({
      ...state,
      focused: true,
    }));
  };

  const onBlur = () => {
    setstate((oldstate) => ({
      ...state,
      focused: false,
    }));
  };

  return (
    <TextInput
      onFocus={onFocus}
      onBlur={onBlur}
      autoFocus
    />
`

### DIFF
--- a/components/TextInput.js
+++ b/components/TextInput.js
@@ -22,12 +22,13 @@ class TextInput extends PureComponent {
 
   handleBlur() {
     this.setState({ isFocused: false });
+    this.props.onBlur()
   }
 
   handleFocus() {
     this.setState({ isFocused: true });
+    this.props.onFocus()
   }
-
   render() {
     const {
       errorMessage,


components/TextInput.js
```diff
@@ -22,12 +22,13 @@ class TextInput extends PureComponent {
 
   handleBlur() {
     this.setState({ isFocused: false });
+    this.props.onBlur()
```
Definitely-Not-Vlad, line 25: Destructure `onBlur` from `this.props` on the first line of the method with a fallback to `null` if it isn't set. This should be followed by a newline. Then check if `onBlur` is truthy and run it:

```
handleBlur() {
  const { onBlur = null } = this.props;

  this.setState({ isFocused: false });
  if (onBlur) {
    onBlur();
  }
}
```

components/TextInput.js
```diff
@@ -22,12 +22,13 @@ class TextInput extends PureComponent {
...
   }
 
   handleFocus() {
     this.setState({ isFocused: true });
+    this.props.onFocus()
```
Definitely-Not-Vlad, line 30: Destructure `onFocus` from `this.props` on the first line of the method with a fallback to `null` if it isn't set. This should be followed by a newline. Then check if `onFocus` is truthy and run it:

```
handleFocus() {
  const { onFocus = null } = this.props;

  this.setState({ isFocused: true });
  if (onFocus) {
    onFocus();
  }
}
```

components/TextInput.js
```diff
@@ -22,12 +22,13 @@ class TextInput extends PureComponent {
...
   }
-
   render() {
```
Definitely-Not-Vlad, line 32: Reintroduce this newline between `handleFocus` and `render`, we separate all methods with a newline.

Definitely-Not-Vlad (changes requested):
Great thinking, somehow it didn't occur to me that people would want to include their own specific `onFocus` and `onBlur`. Just an oversight.

PR #552, opened 2020-12-08
Feature/seext 4198 videos not started in full screen on android devices

tomislav-arambasic:
Android users didn't have the option to view videos in full-screen AT ALL.
I've done the best I could, even tho we've got different Vimeo and Youtube player behaviors now.

- added option (button) to enter / leave full-screen
- adjusted Vimeo to start in full-screen. Couldn't do the same for YT, as this possibility is a part of Vimeo embed url params. YT doesn't support this, or I couldn't find it.

iOS videos always worked perfectly, didn't need these extra Webview params.

Tried to make a video run in landscape full-screen on play button press, but didn't find a way to do it using react-native-webview.
If anyone knows anything about this, please let me know.

Before (no full-screen option):
Screenshot 2020-12-08 at 17 20 58
Screenshot 2020-12-08 at 17 21 07

After video:
https://streamable.com/i02kmd

### DIFF
--- a/components/Video/Video.js
+++ b/components/Video/Video.js
@@ -70,6 +70,8 @@ class Video extends PureComponent {
     return (
       <View style={style.container}>
         <WebView
+          allowsFullscreenVideo
+          mediaPlaybackRequiresUserAction={false}
           style={{width, height}}
           source={createSourceObject(source, playerParams, poster)}
           scrollEnabled={false}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.0.15",
+  "version": "4.0.16",
   "description": "Styleable set of components for React Native applications",
   "dependencies": {
     "@shoutem/animation": "~0.12.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.0.16",
+  "version": "4.0.15",
   "description": "Styleable set of components for React Native applications",
   "dependencies": {
     "@shoutem/animation": "~0.12.4",
--- a/components/Video/VideoSourceReader.js
+++ b/components/Video/VideoSourceReader.js
@@ -30,7 +30,7 @@ function getYouTubeEmbedUrl(id, playerParams) {
 }
 
 function getVimeoEmbedUrl(id) {
-  return `https://player.vimeo.com/video/${id}?title=0&byline=0&portrait=0`;
+  return `https://player.vimeo.com/video/${id}?title=0&byline=0&portrait=0&playsinline=0`;
 }
 
 /**


components/Video/Video.js
```diff
@@ -70,6 +70,8 @@ class Video extends PureComponent {
     return (
       <View style={style.container}>
         <WebView
+          allowsFullscreenVideo
+          mediaPlaybackRequiresUserAction={false}
```
Definitely-Not-Vlad, line 74: Are we sure we want to have `mediaPlaybackRequiresUserAction` set to `false`?

I know everyone hates websites that autoplay videos, I doubt they have a different feeling towards mobile apps.

This would also be a change from our current usage, which isn't outlined anywhere in the issue description.

In my opinion, this shouldn't be changed or set to `false` in general.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.0.15",
+  "version": "4.0.16",
```
Definitely-Not-Vlad, line 3: We don't change the version in feature branches.

Definitely-Not-Vlad (changes requested):
Seeing as this was meant to be done first thing in the morning and Tomislav is on sick leave, I offer to make these changes and release them myself.

components/Video/Video.js
tomislav-arambasic: This is **not** autoplay option, even I understand why you'd think that, I've been there.
When user presses play button on Vimeo player, player goes full-screen. After video loads, player stops and user needs to press play again. This prop "fixed" that. Now, after user presses play in inline player, it goes full-screen, loads and starts playing.

Check "After video" for reference.

toroz: Definitely agree with Vlad that we do not want to autoplay a video as soon as it renders on the screen, but Tomislav did show me the functionality and it doesn't do so. We do need to check whether this "broke" iOS behaviour, though. 

I don't understand why did we need to set it explicitly to false if its value is false by default?

Anyhow, the behavior we're seeking is:
1. when the user presses "play", the video is opened in fullscreen and starts playing
2. user can, at any time, switch to landscape mode (rotate their mobile phone) and back
3. user can, at any time, exit the fullscreen mode and then enter back in

Definitely-Not-Vlad: I'll run it on iOS and check if that's the behavior.

tomislav-arambasic: I'm sorry. Its value is true by default.

iOS worky fine too, same as before. I should've attached iOS video too.
I'll do it soon, so you can see

Definitely-Not-Vlad: I'll take your word for it and accept responsibility if it's wrong.

Merging this and will begin release soon.

Definitely-Not-Vlad (approved):

components/Video/Video.js
tomislav-arambasic: 😱

PR #558, opened 2020-12-11
Disable full screen Vimeo player on Android devices

tomislav-arambasic:
As we (Tara, Goran, Davor, and me) discussed, we'll be disabling Vimeo player full-screen option on Android devices.
If Android users were in Vimeo full-screen portrait mode and used the hardware back button, it'd cause the app to show a blank screen, making the app unusable.

Vimeo player repo has this issue reported and unresolved.
https://github.com/vimeo/player.js/issues/514

So, on Android, we'll be showing a full-screen option for Youtube videos only.
On iOS everything works perfectly well, as all videos are played inside their native player.

Notes:
- Removed `playsinline=0`, as it was before new Video/Vimeo changes. Vimeo player controls are not shown if this is false
- Keeping `mediaPlaybackRequiresUserAction={false}`. Without it, if the user presses play on Vimeo player, the video loads, but it stops again - requiring the user to press play again. With this prop, the video is played straight after load.
- Initialized sourceReader inside class. This way we'll avoid initializing it in every render() call, as it was before.

Android video:
https://streamable.com/g9gtis

iOS video:
https://streamable.com/efxjq0

### DIFF
--- a/components/Video/Video.js
+++ b/components/Video/Video.js
@@ -8,8 +8,7 @@ import { connectStyle } from '@shoutem/theme';
 
 import VideoSourceReader from './VideoSourceReader';
 
-function createSourceObject(source, playerParams, poster) {
-  const sourceReader = new VideoSourceReader(source.uri, playerParams);
+function getSource(sourceReader, poster) {
   const url = sourceReader.getUrl();
 
   if (sourceReader.isEmbeddableVideo()) {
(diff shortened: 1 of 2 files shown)


components/Video/Video.js
```diff
@@ -57,23 +56,34 @@ class Video extends PureComponent {
...
       style,
-      playerParams,
       poster,
     } = this.props;
 
+    // https://github.com/vimeo/player.js/issues/514
+    // Vimeo player crashes the app, if played in full screen portrait mode and if user
+    // tries to navigate back using Android hardware back button. Disableing full screen
+    // option for Vimeo until their player is fixed.
+    const isYoutubeVideo = this.sourceReader.isYouTube;
```
Definitely-Not-Vlad, line 78: Why not add `|| Platform.OS === 'ios'`?

You said it works on iOS without issue, it would be good to then allow iOS devices to display it fullscreen.

If this requires the URL parameter in `VideoSourceReader.js` (`playsinline=0`), then you can add a check for device OS there and choose whether to add the parameter or not.

Definitely-Not-Vlad (changes requested):
Requesting change, but if the comment I made doesn't make sense, I can approve after you reply.

components/Video/Video.js
tomislav-arambasic: iOS plays videos in full-screen by default, on video start, it goes full-screen automatically.

Definitely-Not-Vlad: Perfect, approving.

Definitely-Not-Vlad (approved):

Definitely-Not-Vlad:
Change branch to release if this is meant to go in release.

components/Video/Video.js
tomislav-arambasic: Should've given more details in the PR description again, I guess.

iOS plays both Vimeo and Youtube videos perfectly. It always did.
It's because they're playing videos in a native player.
On iOS, when the user starts the video, the player goes full-screen and plays. Player controls are native.
So, almost none of these params are "needed" by iOS.

On Android devices, videos are played in Vimeo/Youtube players, inside WebView.
Player controls are theirs, all user inputs (including pressing player controls) are handled by the video player itself.
That's why we are disabling Vimeo Android full-screen - Vimeo player has a bug.

Passed `playsinline=0` was added only because we were trying to enable Vimeo full-screen video play in the previous Video component PR.
We haven't had the option to go full-screen before.
So, `playsinline=0`would cause the Vimeo player to go full-screen automatically when the user started playing the video.
Now, that we're "rollbacking" these changes Vimeo player should play inline again, without the option for full-screen, like it was before.
On iOS, this does nothing, it works perfectly itself.

PR #561, opened 2020-12-15
Replace formatLineHeight with the new method

majaklajic:


Definitely-Not-Vlad (approved):
Immaculate. Inspiring. Incredible.

PR #568, opened 2021-01-06
detect iPhone 12, iPhone 12 Pro and iPhone 12 Pro Max

### LINKED ISSUE #566: Navigation bar issue with notch on iPhone 12
Currently we are facing an issue where  some part of the navigation bar is hidden under the notch. Should there be some special handling like for iPhone X and iPhone XR ?

helloagain-dev:
Fixes #566


toroz:
We will publish this change with the next shoutem/ui version (4.0.19).

PR #573, opened 2021-02-12
Adjust ImageGalleryOverlay title padding

tomislav-arambasic:
The title was offset after [Screen padding changes](https://github.com/shoutem/mobile/pull/490).
Verified on Android too, I just didn't take a screenshot.

Screenshot 2021-02-12 at 09 15 58

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.1.0",
+  "version": "4.2.0",
   "description": "Styleable set of components for React Native applications",
   "scripts": {
     "lint": "eslint .",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.1.1",
+  "version": "4.1.0",
   "description": "Styleable set of components for React Native applications",
   "scripts": {
     "lint": "eslint .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "4.1.0",
+  "version": "4.1.1",
```
Definitely-Not-Vlad, line 3: Version changes are made in release branches, so if there's multiple feature branches active, they don't have to worry about which version they have to set.

theme.js
```diff
@@ -2647,15 +2647,8 @@ export default (variables = defaultThemeVariables) => ({
     '.full-screen': {
       title: {
         container: {
-          // We want the title background gradient to be
-          // visible underneath the navigation bar, but the
-          // title text should be rendered below the
-          // navigation bar.
```
Definitely-Not-Vlad, line 2653: Did you verify this background gradient behavior after your changes or just the `ImageGalleryOverlay`?

Definitely-Not-Vlad (changes requested):

theme.js
tomislav-arambasic: Honestly, I've seen no background gradient, ever. This seems like an old comment, while we had a different design.
I'll check it again to verify, I'm curious now

tomislav-arambasic: I've reverted deleting this comment.
I see some styles I can't figure out, including gradient.
I see we're setting some background gradient for title and description, but it's not visible in my app.
I'll leave it as is, just in case.

Definitely-Not-Vlad (approved):

PR #576, opened 2021-02-23
Feature/iedc additions

sstimac:
IEDC Generic components we can make use of

- TabMenu - horizontal tabular menu that could be used as a replacement for DropDownMenu ( can be upgraded with more animated transitions )
- InlineDropDownMenu - Collapsible field with animated item transition. Ideal for forms or other places where you want collapsible list of options to choose from
- YearPicker - Combination of a button and appropriate modal, allowing you to make a multiple year selection choice. Very similar to calendar day picker

In addition to this, there are some new styles for our standard set of components, as well as common style pattern inclusions, like adding gutter variances to touchableOpacity , and similar


components/InlineDropDownMenu/InlineDropDownMenu.js
```diff
@@ -0,0 +1,130 @@
...
+      options,
+    } = this.props;
+    const { collapsed } = this.state;
+
+    return (
+      <View>
+        <TouchableOpacity style={style.container} onPress={this.handleToggleMenuPress}>
+          <Caption styleName="muted md-gutter-bottom">{heading}</Caption>
+          <View styleName="space-between horizontal v-center">
+            <Text>{selectedOption?.title}</Text>
```
Definitely-Not-Vlad, line 98: `selectedOption?.title` Smexy.

Although, won't this result in a Text component being rendered regardless of there being a title or not? Is this necessary to maintain the layout structure and is it inexpensive to rendered the Text regardless?

Food for thought, I don't mind either way, personally I don't think it's a big deal if an extra Text component is rendered in the unlikely case it has no title.

components/InlineDropDownMenu/InlineDropDownMenuItem.js
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <AnimatedTouchable
+        style={[
+          style.container,
+          {
+            transform: [
+              {
+                translateX: this.animatedValue.interpolate({
+                  inputRange: [0, 1],
+                  outputRange: [window.width, 0]
```
Definitely-Not-Vlad, line 66: Trailing commas? I assume prettier was just doing what the config from @shoutem/ui was telling it to, should we look into that?

components/TabMenu/TabMenu.js
```diff
@@ -0,0 +1,64 @@
+import PropTypes from 'prop-types';
+import autoBindReact from 'auto-bind/react';
+import React, { PureComponent } from 'react';
```
Definitely-Not-Vlad, line 3: Imports from `'react'` should be first.

components/TabMenu/TabMenu.js
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    const { style, options } = this.props;
+
+    return (
+      <ScrollView
+        horizontal
+        contentContainerStyle={style.container}
+        showsHorizontalScrollIndicator={false}
+        style={{ flexGrow: 0 }}
```
Definitely-Not-Vlad, line 54: Should this be hardcoded or provided through theme regardless of being the _required_ behavior?

Just a question, not requesting changes.

components/YearPicker/YearPickerButton.js
```diff
@@ -0,0 +1,82 @@
...
+          {tooltip}
+        </Text>
+        <AnimatedIcon
+          name="drop-down"
+          style={{
+            ...style.icon,
+            transform: [{
+              rotate: this.dropDownIconValue.interpolate({
+                inputRange: [0, 1],
+                outputRange: ['0deg', '180deg']
```
Definitely-Not-Vlad, line 72: Trailing commas? Check other comment with same "issue".

theme.js
```diff
@@ -2863,4 +2890,126 @@ export default (variables = defaultThemeVariables) => ({
...
+    },
+    icon: {
+      color: variables.text.color,
+    }
+  },
+
+  'shoutem.ui.YearPickerModal': {
+    container: {
+      padding: 8,
+      backgroundColor: '#F9F9F9',
```
Definitely-Not-Vlad, line 2967: Should we maybe use colors that can be modified via theme variables i.e. paper color, background color, etc.? Rather than hardcoding and then any changes wanted require your custom theme to directly modify `shoutem.ui.YearPickerModal` instead of just syncing up with your theme's `paperColor`?

Let me know if my question doesn't make sense, I'm not sure I worded it the best.

Definitely-Not-Vlad (changes requested):
Requesting changes only to get answers to my questions, I don't actually think any changes need to be made.

One overarching question that I have is: can the year-picker become whatever-picker? Because the idea seems great, but if I'm understanding the way it works correctly, I don't see a reason it should be just years.

Let me know if I'm misunderstanding something and great job with these additions, love to see stuff like this being done!

toroz (changes requested):

components/InlineDropDownMenu/InlineDropDownMenu.js
sstimac: Yes it will. that being said ,we don't want to handle that case. PropTypes should display proper warning if the required shape isn't present

components/InlineDropDownMenu/InlineDropDownMenuItem.js
sstimac: I'm all up for copying lint configs to ui ( that we have in mobile repo ) if we don't have them already

sstimac:
> Requesting changes only to get answers to my questions, I don't actually think any changes need to be made.
> 
> One overarching question that I have is: can the year-picker become whatever-picker? Because the idea seems great, but if I'm understanding the way it works correctly, I don't see a reason it should be just years.
> 
> Let me know if I'm misunderstanding something and great job with these additions, love to see stuff like this being done!

Yes, i'd very much like to upgrade this like you described, at some point. Due to the time constraint i decided to write a simpler logic for the time being, as working with years ( numbers ) simplifies the logic quite a bit.

Definitely-Not-Vlad (approved):
LGTM.

toroz (approved):

PR #579, opened 2021-03-16
Add Wiggle animation option to TextInput

tomislav-arambasic:
https://user-images.githubusercontent.com/57353746/111321820-95a0db80-8668-11eb-8fa8-3b290cd85167.mov

### DIFF
--- a/components/TextInput.js
+++ b/components/TextInput.js
@@ -1,11 +1,11 @@
-import React, { PureComponent } from 'react';
-import autoBindReact from 'auto-bind/react';
-import PropTypes from 'prop-types';
-import { TextInput as RNTextInput } from 'react-native';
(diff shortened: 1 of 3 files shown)


components/TextInput.js
```diff
@@ -27,7 +27,13 @@ class TextInput extends PureComponent {
   }
 
   render() {
-    const { errorMessage, highlightOnFocus, style, ...otherProps } = this.props;
+    const {
+      errorMessage,
+      highlightOnFocus,
+      startErrorAnimation,
```
sstimac, line 33: i don't like the new prop for this. It should be tied to the errorMessage prop somehow

sstimac (changes requested):

sstimac (approved):

PR #586, opened 2021-04-03
Fix checkbox rectangle off/on loading

Eraz1997:
Update components/Icon/assets/index.js to load a couple
of checkbox SVGs correctly. Extension missing, app breaks
when loading <Examples/> on a brand new project.
Fix names.


Definitely-Not-Vlad (approved):

Definitely-Not-Vlad:
Thanks for this, we've got a release branch open so I'll close this as we've added those changes there.

PR #587, opened 2021-04-07
Update index.js

Ninjinka:
missing extension on lines


Definitely-Not-Vlad (approved):
Thanks so much for catching this.

I'll merge this so it goes out with the next first next release we have.

PR #591, opened 2021-04-29
Feature/seext 9738 fix faulty behaviour of br tags

tomislav-arambasic:


Definitely-Not-Vlad (approved):
LGTM code wise, but just to make sure, this would be the expected behavior for the user? If yes, good to go, merge away at your discretion, otherwise, se if there is a way to get the expected behavior.

Hope that makes sense.

html/components/SimpleHtml.js
```diff
@@ -71,6 +71,32 @@ class SimpleHtml extends PureComponent {
...
+   * <div>first line<br></div>
+   * <div>second line<br></div>
+   *
+   * If br is left as is, <Html> will show empty line under div, not just
+   * break into new line
+   */
+  alterChildren(node) {
+    const { children, name } = node;
+
+    if ((name === 'div' || name === 'p') && children && children.length > 1) {
```
toroz, line 87: Only for `div` & `p`?

tomislav-arambasic: Headings have top and bottom "margins" in editor & on mobile, so they're good.
Everything looks the same, on editor & mobile now.

tomislav-arambasic:
> LGTM code wise, but just to make sure, this would be the expected behavior for the user? If yes, good to go, merge away at your discretion, otherwise, se if there is a way to get the expected behavior.
> 
> Hope that makes sense.

I wasn't sure so I checked.
When I pasted the same HTML code on 2 other HTML interpreter sites, I've seen that one `<br>`, at the end of the line, doesn't create empty line below, but only breaks the text to new line.
The other case, when there is a `div` or `p` tag, and it has no other children but `br`, then it creates a new line.

`Html` component creates empty line when it sees `<br>`.
So I handled scenario when there is other children in `div` or `p`, other than br.

Screenshot 2021-04-29 at 13 04 49

html/components/SimpleHtml.js
tomislav-arambasic: Not sure if you get notified on tagging you when I edit the comment, so I'll leave this here :P 

image

PR #600, opened 2021-06-07
Fix .svg files

Definitely-Not-Vlad:
Current .svg where:
- "on" icon checkmark is always white
- "on" icon is always black (instead of taking the color from theme icon tint variable)
- black border exists
Screenshot 2021-06-07 at 14 08 18

New .svg where:
- "on" icon checkmark is transparent
- "on" icon uses theme icon tint variable
- no border exists
Screenshot 2021-06-07 at 14 06 24


majaklajic (approved):
Good one!

toroz (approved):
👏

PR #609, opened 2021-06-15
Feature/add grid row no padding and extend list view with index

tomislav-arambasic:
`GridRow` component has padding top & right - 5px.
Added `no-padding` option for it.

Passing item index to `ListView`'s `renderRow` now.

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -1430,6 +1430,10 @@ export default (variables = defaultThemeVariables) => ({
     justifyContent: 'space-between',
     paddingRight: variables.smallGutter,
     paddingTop: variables.smallGutter,
+
+    '.no-padding': {
+      padding: 0,
+    },
   },
 
   //
--- a/theme.js
+++ b/theme.js
@@ -1424,16 +1424,16 @@ export default (variables = defaultThemeVariables) => ({
       marginBottom: 0,
     },
 
+    '.no-padding': {
+      padding: 0,
+    },
+
     flex: 1,
     flexDirection: 'row',
     alignItems: 'flex-start',
     justifyContent: 'space-between',
     paddingRight: variables.smallGutter,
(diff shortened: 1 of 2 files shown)


theme.js
```diff
@@ -1430,6 +1430,10 @@ export default (variables = defaultThemeVariables) => ({
     justifyContent: 'space-between',
     paddingRight: variables.smallGutter,
     paddingTop: variables.smallGutter,
+
+    '.no-padding': {
+      padding: 0,
+    },
```
Definitely-Not-Vlad, line 1436: When we write styling in theme files, we usually put `styleName`s before any direct style.

What I mean to say with this is that this should be after `'*'` but before `flex: 1,` in the definition for `'shoutem.ui.GridRow'`.

Definitely-Not-Vlad (changes requested):

theme.js
tomislav-arambasic: Forgot to point it out - no-padding styleName is overwriting direct style, paddingRight & paddingTop are defined in direct style

Definitely-Not-Vlad: Oh so it won't work if defined before the directly defined style?

Did you try it out?

Definitely-Not-Vlad: Looking at [Screen component styling here](https://github.com/shoutem/ui/blob/develop/theme.js#L896), `'.paper'` is defined before `backgroundColor: variables.backgroundColor`, yet it still overrides it.

tomislav-arambasic: Yep, you're right.
I was pretty sure it won't work, check now, works as you said. Thanks

Definitely-Not-Vlad (approved):
If you haven't already, please double check that everything is in order with the styleName you added.

PR #615, opened 2021-07-05
Feature/datetime picker

tomislav-arambasic:
On iOS, DateTimePicker is presented as "standalone component", it has no native background or any other UI.
That's why we define modal & place it inside, with confirm button.

On Android, there's no need for additional UI, it's handled natively.

Android:
Screenshot 2021-07-16 at 15 20 51
Screenshot 2021-07-16 at 15 20 46

iOS:
Screenshot 2021-07-16 at 15 22 38
Screenshot 2021-07-16 at 15 22 32

### DIFF
--- a/components/DateTimePicker.js
+++ b/components/DateTimePicker.js
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react';
 import RNCDateTimePicker from '@react-native-community/datetimepicker';
 import autoBindReact from 'auto-bind/react';
+import PropTypes from 'prop-types';
 import { Platform } from 'react-native';
 import Modal from 'react-native-modal';
 import { connectStyle } from '@shoutem/theme';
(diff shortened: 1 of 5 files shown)


components/DateTimePicker.js
```diff
@@ -0,0 +1,161 @@
...
+import autoBindReact from 'auto-bind/react';
+import PropTypes from 'prop-types';
+import { Platform } from 'react-native';
+import Modal from 'react-native-modal';
+import { connectStyle } from '@shoutem/theme';
+import { Icon } from './Icon';
+import { Text } from './Text';
+import { TouchableOpacity } from './TouchableOpacity';
+import { View } from './View';
+import { Button } from './Button';
```
majaklajic, line 12: Please order imports alphabetically.

majaklajic (changes requested):
Looks perfect, just one small CR related to imports.

components/DateTimePicker.js
```diff
@@ -0,0 +1,161 @@
...
+      return this.setState({ value });
+    }
+
+    if (event.type === 'dismissed') {
+      return this.setState({ showPicker: false });
+    }
+
+    const { onValueChanged } = this.props;
+
+    this.setState({ showPicker: false });
```
Definitely-Not-Vlad, line 47: Use `this.handleHidePicker()` like you do elsewhere in the code.

theme.js
```diff
@@ -3055,4 +3060,20 @@ export default (variables = defaultThemeVariables) => ({
...
     },
   },
+
+  'shoutem.ui.DateTimePicker': {
+    buttonContainer: {
+      width: 50,
+      height: 50,
+      backgroundColor: '#222222',
+    },
+    icon: { color: '#FFFFFF', height: 30, width: 30 },
```
Definitely-Not-Vlad, line 3070: AFAIK `Icon` receives color via `fill={color}`, not via style prop.

I've been writing it as:
`iconFill: '#FFFFFF'`

Then using that in the icon as `fill={style.iconFill}`.

If there's multiple icons, I gave the style properties different names, e.g. `confirmIcon` and `confirmIconFill`.

Definitely-Not-Vlad (changes requested):

components/DateTimePicker.js
```diff
@@ -0,0 +1,161 @@
...
+        </TouchableOpacity>
+        <TouchableOpacity
+          onPress={this.handleShowPicker}
+          styleName="h-center v-center"
+          style={style.buttonContainer}
+        >
+          <Icon name="drop-down" style={style.icon} />
+        </TouchableOpacity>
+        {isIos && (
+          <Modal
```
ltatarev, line 95: Not a request change, but personally I like to add props such as `swipeDirection` and `onSwipeComplete` to modal to allow users dismissing modal via swiping up or down. 
This is more of a UX change so it's up to your preference.

theme.js
tomislav-arambasic: Doesn't work for me.
It's because `Icon` component from `shoutem/ui` sets `style.color` as `fill` param.

components/DateTimePicker.js
tomislav-arambasic: This looks super cool, defo UX improvement :D 
Thx!

https://user-images.githubusercontent.com/57353746/126150762-ba44f37b-290d-4e1b-8001-2378b10507d0.mov

Definitely-Not-Vlad (approved):

ltatarev (approved):

majaklajic (approved):

PR #617, opened 2021-07-13
Feature/navigation compatibility

sstimac:
Pr containing changes for the new navigation ( mobile repo changes ).
The changes handle the removal of navigationBar component that was used to style navbar. Those styles were optimised to follow react-navigation ways of styling ( each navbar piece has it's separate styling ), and to reduce unnecessary code that was introduced due to the inheritance way of styling on the prior platform


Definitely-Not-Vlad:
Before I review, could you please elaborate on the changes in theme.js? I'm having a hard time reading the diff.

What did you do and why?

majaklajic:
@sstimac Maybe a stupid question, but are we sure we're not gonna break anything in Rubicon or other extensions by removing these chunks of code?

Goran1 (approved):

PR #630, opened 2021-08-30
Remove flex style prop for SearchField input

Definitely-Not-Vlad:
This was causing issues with the text being entered not being visible at all.

Don't ask how, I wouldn't be able to explain.

I checked every implementation I could think of and didn't find any adverse effects.


majaklajic (approved):
Looks good.

tomislav-arambasic (approved):

PR #637, opened 2021-09-06
Feature/adjust unsplash images app style

tomislav-arambasic:
This adds styling for Unsplash images, inserted into RTE via builder - calculating max width of image, adding margin top & bottom & centering the description below image.
Height is auto-calculated, keeping image ratio.

https://user-images.githubusercontent.com/57353746/132230844-51a1812d-7333-4d90-821f-bf10237cd331.mov

### DIFF
--- a/html/components/SimpleHtml.js
+++ b/html/components/SimpleHtml.js
@@ -25,6 +25,24 @@ import getEmptyObjectKeys from '../services/getEmptyObjectKeys';
 import isValidVideoFormat from '../services/isValidVideoFormat';
 import Image from './Image';
 
+const UNSPLASH_IMAGE_CONTAINER_ID = 'unsplash-image-container';
+
+function alterUnsplashImagesNode(node) {
+  const styleAttrib = _.get(node, 'attribs.style', '').trim();
+
+  if (node.name === 'div' && node.attribs.id === UNSPLASH_IMAGE_CONTAINER_ID) {
+    const imageContainerWithMargins = {
+      attribs: {
+        style: `${styleAttrib}margin-top: 15px; margin-bottom: 30px;`,
+      },
+    };
+    const resolvedNode = _.merge(node, imageContainerWithMargins);
+    return resolvedNode;
+  }
+
+  return null;
+}
+
 class SimpleHtml extends PureComponent {
   static propTypes = {
     body: PropTypes.string,
@@ -60,6 +78,10 @@ class SimpleHtml extends PureComponent {
 
     const styleAttrib = _.get(node, 'attribs.style', '').trim();
 
+    if (node.attribs?.id === UNSPLASH_IMAGE_CONTAINER_ID) {
+      return alterUnsplashImagesNode(node);
+    }
+
     if (node.name === 'table') {
       return tableAlterNode(node);
     }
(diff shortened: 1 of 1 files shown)


html/components/SimpleHtml.js
```diff
@@ -25,6 +25,24 @@ import getEmptyObjectKeys from '../services/getEmptyObjectKeys';
 import isValidVideoFormat from '../services/isValidVideoFormat';
 import Image from './Image';
 
+const UNSPLASH_IMAGE_CONTAINER_ID = 'unsplash-image-container';
```
Definitely-Not-Vlad, line 28: This ID, is it something that the RTE on the Builder generates?

If so, I'm not 100% certain this code is meant to be in the @shoutem/ui lib, because the lib itself is independent from the Builder.

Might be better to expose a prop with which we could then handle this. I leave this to your discretion however.

html/components/SimpleHtml.js
```diff
@@ -25,6 +25,24 @@ import getEmptyObjectKeys from '../services/getEmptyObjectKeys';
...
+
+function alterUnsplashImagesNode(node) {
+  const styleAttrib = _.get(node, 'attribs.style', '').trim();
+
+  if (node.name === 'div' && node.attribs.id === UNSPLASH_IMAGE_CONTAINER_ID) {
+    const imageContainerWithMargins = {
+      attribs: {
+        style: `${styleAttrib}margin-top: 15px; margin-bottom: 30px;`,
```
Definitely-Not-Vlad, line 36: What happens if `styleAttrib` already contains `margin-top` or `margin-bottom`? Is this a possible situation?

html/components/SimpleHtml.js
```diff
@@ -25,6 +25,24 @@ import getEmptyObjectKeys from '../services/getEmptyObjectKeys';
...
+      },
+    };
+    const resolvedNode = _.merge(node, imageContainerWithMargins);
+    return resolvedNode;
```
Definitely-Not-Vlad, line 40: I can't recall anymore for certain, but I believe we agreed to do spacing before `return` statements.

html/components/SimpleHtml.js
```diff
@@ -25,6 +25,24 @@ import getEmptyObjectKeys from '../services/getEmptyObjectKeys';
 import isValidVideoFormat from '../services/isValidVideoFormat';
 import Image from './Image';
 
+const UNSPLASH_IMAGE_CONTAINER_ID = 'unsplash-image-container';
+
+function alterUnsplashImagesNode(node) {
+  const styleAttrib = _.get(node, 'attribs.style', '').trim();
+
+  if (node.name === 'div' && node.attribs.id === UNSPLASH_IMAGE_CONTAINER_ID) {
```
Definitely-Not-Vlad, line 33: Since `alterUnsplashImagesNode` is only called when `node.attribs.id === UNSPLASH_IMAGE_CONTAINER_ID`, isn't this if statement redundant? Maybe just check for `node.name === 'div'`?

Leaving to your discretion, maybe I missed something.

Definitely-Not-Vlad (changes requested):

html/components/SimpleHtml.js
tomislav-arambasic: This ID comes from HTML "generated" when inserting Unsplash image - new upcoming feature.
By "generated" I mean I've hardtyped HTML code to be inserted in CMS's Rich Text Editor.

html/components/SimpleHtml.js
tomislav-arambasic: I'll check it out.
I'm 99,9% sure it'd work, because it'd read last margin-top & margin-bottom properties, but still... I'll check.

html/components/SimpleHtml.js
tomislav-arambasic: Oopsie, ty

html/components/SimpleHtml.js
tomislav-arambasic: You're right

html/components/SimpleHtml.js
Definitely-Not-Vlad: Maybe the alternative then is to allow 3rd party developers to turn this alter off?

This might be a separate improvement in and of itself, as we have multiple alters like that, so we may want to give 3rd party developers the chance to override this if we already don't.

For now, let's leave this as it is and revisit this at some point.

tomislav-arambasic:
Closing this as it was unnecessary change. Images are centered by default after all....

PR #638, opened 2021-09-07
feature/SEEXT-10408 Create SimpleHtml's prop for declaring custom alterNode function

tomislav-arambasic:
This should enable us to alter nodes from outside of `SimpleHtml`, from a component rendering `SimpleHtml`, making our life easier & keeping custom logic outside `@shoutem/ui`.

**Before, no alterations:**
Screenshot 2021-09-07 at 09 20 13

Example code for altering `div` node with `unsplash-image-container` ID, defined in `AboutScreen.js`:

```
  alterUnsplashImagesNode(node) {
    const styleAttrib = _.get(node, 'attribs.style', '').trim();

    if (node.attribs?.id === 'unsplash-image-container') {
      const imageContainerWithMargins = {
        attribs: {
          style: `${styleAttrib}margin-top: 115px; margin-bottom: 130px;`,
        },
      };
      const resolvedNode = _.merge(node, imageContainerWithMargins);
      return resolvedNode;
    }

    return null;
  }

  renderInfo(profile) {
    if (!_.get(profile, 'info')) {
      return null;
    }

    return (
      <SimpleHtml
        body={profile?.info}
        customAlterNode={this.alterUnsplashImagesNode}
      />
    );
  }
```

**After, with `customAlterNode`, adding margins:**
Screenshot 2021-09-07 at 09 19 50

### DIFF
--- a/html/components/SimpleHtml.js
+++ b/html/components/SimpleHtml.js
@@ -62,7 +62,15 @@ class SimpleHtml extends PureComponent {
     const styleAttrib = _.get(node, 'attribs.style', '').trim();
 
     if (customAlterNode && _.isFunction(customAlterNode)) {
-      return customAlterNode(node, cssObjectToString, cssStringToObject);
+      const resolvedNode = customAlterNode(
+        node,
+        cssObjectToString,
+        cssStringToObject,
+      );
+
+      if (resolvedNode) {
+        return resolvedNode;
+      }
     }
 
     if (node.name === 'table') {
--- a/html/components/SimpleHtml.js
+++ b/html/components/SimpleHtml.js
@@ -33,6 +33,7 @@ class SimpleHtml extends PureComponent {
     customTagStyles: PropTypes.object,
     customHandleLinkPress: PropTypes.func,
     unsupportedVideoFormatMessage: PropTypes.string,
+    customAlterNode: PropTypes.func,
   };
 
   constructor(props) {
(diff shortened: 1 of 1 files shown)


html/components/SimpleHtml.js
```diff
@@ -56,10 +57,14 @@ class SimpleHtml extends PureComponent {
...
   alterNode(node) {
-    const { style } = this.props;
+    const { customAlterNode, style } = this.props;
 
     const styleAttrib = _.get(node, 'attribs.style', '').trim();
 
+    if (customAlterNode && _.isFunction(customAlterNode)) {
+      return customAlterNode(node, cssObjectToString, cssStringToObject);
+    }
+
```
Definitely-Not-Vlad, line 67: If we pass `customAlterNode` for `img` tags, won't this prevent our current existing alter nodes for table and others to be used?

tomislav-arambasic: It will, but...
The idea is that you point to specific node IDs & alter only those.
If you point to specific ID only, inside your `customAlterNode` fn, other `img` tags will be altered by `SimpleHtml`.

tomislav-arambasic: Ah, I see what I've done wrong...
It'll return `null` too early.

Definitely-Not-Vlad: I understand that, but `alterNode` is going to check if `customAlterNode` exists, if it does, it will execute it and that's it, return stops function execution, it doesn't matter if `customAlterNode` returns `null`, no? Because there's no way `alterNode` can know the ID `customAlterNode` is looking for, it'll just always execute it if it exists for every tag, preventing the other alter functions inside `alterNode` to be executed.

Maybe I'm not getting something but this looks like it just won't ever do anything other than `customAlterNode` when it's defined.

Definitely-Not-Vlad (approved):

html/components/SimpleHtml.js
tomislav-arambasic: I was **nedorečen** again, gotta stop doing it.

I assume my last commit is the solution to your question 'coz you approved it.
By "too early null return" - I meant exactly what you described in comment before.
I've added additional IF statement, if `customAlterNode` returns altered node, it'll return.
If not, if the result of `customAlterNode` is null, it'll continue with the `alterNode` function execution.

PR #652, opened 2021-11-08
Update DateTimePicker

ltatarev:
Updated DateTimePicker component with following: 
(screenshots below)
- added `Cancel` button to iOS modal (Android modal has it by default, so it made sense to add it since modal has no "clear" way of dismissing - other than swipe)
- added better handling for `date` and `datetime` display mode on iOS
- added `datetime` ability on Android: first open date modal, and after open time modal

<table>
<tr>
<td>

</td>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
 </table>

### DIFF
--- a/components/DateTimePicker.js
+++ b/components/DateTimePicker.js
@@ -13,19 +13,40 @@ import { View } from './View';
 
 const isIos = Platform.OS === 'ios';
 
+const MODES = { DATE: 'date', DATETIME: 'datetime', TIME: 'time' };
+
+const DISPLAY_MODES = {
+  [MODES.DATE]: 'inline',
+  [MODES.DATETIME]: 'inline',
+  [MODES.TIME]: 'spinner',
+};
+
+function getDisplayMode(mode) {
+  if (!isIos || !Object.keys(DISPLAY_MODES).includes(mode)) {
+    return 'default';
+  }
+
+  return DISPLAY_MODES[mode];
+}
+
 class DateTimePicker extends PureComponent {
   constructor(props) {
     super(props);
 
     autoBindReact(this);
 
+    const { mode, value } = props;
+
     this.state = {
+      display: getDisplayMode(mode),
       showPicker: false,
-      value: props.value,
+      showTimePicker: false,
+      value,
     };
   }
 
   componentDidUpdate() {
+    // todo: check why set state here
     if (!isIos) {
       const { value } = this.props;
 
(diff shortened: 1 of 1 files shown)


components/DateTimePicker.js
```diff
@@ -128,29 +179,41 @@ class DateTimePicker extends PureComponent {
...
     );
   }
 }
 
 DateTimePicker.propTypes = {
+  cancelButtonText: PropTypes.string,
   confirmButtonText: PropTypes.string,
   is24Hour: PropTypes.bool,
-  mode: PropTypes.string,
+  mode: PropTypes.oneOf(Object.values(MODES)),
```
Definitely-Not-Vlad, line 209: I like that a lot!

Definitely-Not-Vlad (approved):
LGTM!

components/DateTimePicker.js
```diff
@@ -13,19 +13,40 @@ import { View } from './View';
...
+      display: getDisplayMode(mode),
       showPicker: false,
-      value: props.value,
+      showTimePicker: false,
+      value,
     };
   }
 
   componentDidUpdate() {
+    // todo: check why set state here
```
sstimac, line 49: there was some buggy behaviour with this in the past, which is why we needed this piece of dirt. You can check it out and remove if possible ( it was android only issue )

components/DateTimePicker.js
```diff
@@ -13,19 +13,40 @@ import { View } from './View';
 
 const isIos = Platform.OS === 'ios';
 
+const MODES = { DATE: 'date', DATETIME: 'datetime', TIME: 'time' };
```
sstimac, line 16: export this for outside usage / manipulation

sstimac (approved):
very nice !

components/DateTimePicker.js
ltatarev: In my example app everything worked fine on Android, but since I'm not familiar with all use cases of this component, I would maybe keep it here for now

components/DateTimePicker.js
ltatarev: 👍
Will rename to `DATEPICKER_MODES` for better readability when importing

components/DateTimePicker.js
tomislav-arambasic: Let's do 1 more check before merging this - switch simulator/device to dark mode & test this.
I remember there was some weird behavior that was happening only if device had dark mode/theme (other than picker not being visible)

tomislav-arambasic (changes requested):
I'm requesting changes so that you test one more time before deleting "hack" applied.
Test on device with dark mode & see what happens if we delete the `componentDidUpdate` lifecycle

ltatarev:
> I'm requesting changes so that you test one more time before deleting "hack" applied. Test on device with dark mode & see what happens if we delete the `componentDidUpdate` lifecycle

Tested on Samsung S9 and iPhone 11 Pro Max with both light and dark mode, everything worked fine after deleting `componentDidUpdate`. My guess is that the hack may night be related to dark modes, so I'm not confident with deleting it for now.

tomislav-arambasic (approved):

PR #653, opened 2021-11-16
Feature: Scoped theming support

sstimac:
Expose resolver service and it's default instance ( Retain backwards compatibility as well as standalone UI usage )

- default UI theme variables are still present and resolveable via the new service
- when plugged into Shoutem lifecycle, shoutem.theme scope is present, overriding the default UI variables ( if the vars are actually present )
- We retain backwards compatibility, meaning that everything works as it used to, until we start adding theme scopes. at that point, we will only need to edit Rubicon , including the proper variable scopes inside the resolution flow


services/variableResolver.js
```diff
@@ -0,0 +1,57 @@
...
+    );
+  }
+
+  // Currently supports 2 function signture
+  // resolveVariable(variableName) -> Tries to get the variable under the shoutem.theme namespace
+  // resolveVariable(scope, variableName) -> Tries to get the variable under specific scope
+  // Both signatures default to root parameter for the required variable name
+  // Variable name can be stringified object path or a regular string, i.e -> text.color
+  resolveVariable(...params) {
+    if (arguments.length === 1) {
```
Definitely-Not-Vlad, line 39: `params` instead of `arguments`?

Or am I missing something?

Definitely-Not-Vlad: Also, I'm not certain I understand why arguments are passed as `...params` or why `params.length` is used, would it not be better to receive two arguments if we only expect two arguments? What am I missing?

services/variableResolver.js
```diff
@@ -0,0 +1,57 @@
...
+    const variablePath = _.split(variable, '.');
+
+    return _.get(
+      this.variables,
+      [scope, ...variablePath],
+      _.get(this.variables, variablePath),
+    );
+  }
+
+  // Currently supports 2 function signture
```
Definitely-Not-Vlad, line 33: signature*

services/variableResolver.js
```diff
@@ -0,0 +1,57 @@
...
+  // resolveVariable(scope, variableName) -> Tries to get the variable under specific scope
+  // Both signatures default to root parameter for the required variable name
+  // Variable name can be stringified object path or a regular string, i.e -> text.color
+  resolveVariable(...params) {
+    if (arguments.length === 1) {
+      return this.getScopedVariable(THEME_EXTENSION_SCOPE, params[0]);
+    }
+
+    if (arguments.length === 2) {
+      return this.getScopedVariable(params[0], params[0]);
```
Definitely-Not-Vlad, line 44: You pass `params[0]` twice, is this intentional? I assume it isn't.

Definitely-Not-Vlad (changes requested):

services/variableResolver.js
sstimac: lint complains about arguments syntax, so that's why the spread.

Reasoning behind is the fact that we want to retain simple usage for UI users outside shoutem. For them, there are no scopes unless they want them to exist by defining custom variables. So, the method signature remains simple resolveVariable(variableName).

Other use case is shoutem development scenario, where we target the scoped variable, so we use resolveVariable(scope,variableName)

In both cases, we default to variables contained in the root of the vars object. This is just way less confusing for people using the standalone package.

Another argument is less abstraction for the custom devs. For them, there are general variables ( theme variables ), and scoped ones. Under the hood, general variables are also scoped, but there's no reason for the dev to be aware of that ( simpler interface ).

let me know if i make sense :D

services/variableResolver.js
sstimac: no it is not , lol :D

services/variableResolver.js
Definitely-Not-Vlad: Why not use `resolveVariable(variableName, scope)` then? If `scope` exists, you apply it, otherwise you just use `resolveVariable(variableName)`.

I don't understand what eslint would complain about.

Sorry, something probably just isn't clicking in my head regarding this, I just don't see the need for using the param spread or checking length of params

sstimac: linter explicitly told me to use spread instead of arguments syntax... He doesn't like it i guess.

What you suggesting was my initial approach, but i wanted the scope to go first. It's more natural somehow to me( probably because we follow the actual object structure ). 

`resolveVariable('shoutem.navigation', 'backgroundColor')`. We can change it to what you're suggesting if this doesn't look better to you guys

Definitely-Not-Vlad: I see what you mean, I think it's more important to keep the scope first than it is to have one function defined without spread params for the sake of not having spread params.

Definitely-Not-Vlad (approved):

majaklajic (approved):

services/variableResolver.js
```diff
@@ -0,0 +1,85 @@
...
+    };
+  }
+}
+
+// Expose both, the resolver service, and the default resolver working with
+// shoutem default theme. This way the logic can be used for multiple themes/resolvers, or
+// if needed, you can override the default theme variables using the setVariables method
+export const defaultResolver = new ThemeVariableResolver();
+export const resolveVariable = defaultResolver.resolveVariable;
+export const createScopedResolver = defaultResolver.createScopedResolver;
```
toroz, line 85: Why the explicit export? Just wondering.

toroz (approved):

PR #667, opened 2021-12-22
Add customOnScroll prop and pass it to ScrollDriver properly

Definitely-Not-Vlad:
Adds new prop `customOnScroll`, which will be executed alongside the animations that `@shoutem/animation` applies to `ScrollView`, e.g. `solidify` for the navigation header.

Has to go out _after_ [this PR on @shoutem/ui](https://github.com/shoutem/animation/pull/53).


components/ScrollView/ScrollDriverProvider.js
```diff
@@ -22,6 +22,7 @@ export class ScrollDriverProvider extends PureComponent {
   static propTypes = {
     children: PropTypes.node,
     driver: DriverShape,
+    customOnScroll: PropTypes.func,
```
sstimac, line 25: just onScroll

sstimac (approved):
check the comment from the other PR and adapt the config if necessary

PR #668, opened 2021-12-27
Add inProgress to Button component

tomislav-arambasic:
This is more of an suggestion, let me know if you think we should implement this or not.
In the past few weeks, I've been adding custom loading to buttons a few times, so I thought this might be handy addition.

- added spinner indicator if in progress
- backward compatible
- tried to refactor code a bit

Video:
https://user-images.githubusercontent.com/57353746/147471773-f748d30f-2040-4e61-bb30-e8402b2714e8.mov

### DIFF
--- a/components/Button.js
+++ b/components/Button.js
@@ -5,8 +5,13 @@ import { connectStyle } from '@shoutem/theme';
 import { Spinner } from './Spinner';
 
 class Button extends PureComponent {
-  componentDidUpdate() {
-    LayoutAnimation.easeInEaseOut();
+  componentDidUpdate(prevProps) {
+    const { inProgress: prevInProgress } = prevProps;
+    const { inProgress } = this.props;
+
+    if (prevInProgress !== inProgress) {
+      LayoutAnimation.easeInEaseOut();
+    }
   }
 
   resolveStyle() {
--- a/components/Button.js
+++ b/components/Button.js
@@ -1,29 +1,60 @@
 import React, { PureComponent } from 'react';
-import { TouchableOpacity } from 'react-native';
+import { LayoutAnimation, TouchableOpacity } from 'react-native';
 import { connectAnimation } from '@shoutem/animation';
 import { connectStyle } from '@shoutem/theme';
+import { Spinner } from './Spinner';
 
 class Button extends PureComponent {
-  render() {
-    // The underlayColor is not a valid RN style
-    // property, so we have to unset it here.
-    const style = {
-      ...this.props.style,
+  componentDidUpdate() {
+    LayoutAnimation.easeInEaseOut();
+  }
+
+  resolveStyle() {
+    const { inProgress, style } = this.props;
+
+    if (!inProgress) {
+      return style;
+    }
+
+    return {
+      ...style,
+      ...style?.inProgress,
+      borderRadius: style?.inProgress?.borderRadius || 50,
+      minWidth: style?.inProgress?.minWidth || 30,
+      width: style?.inProgress?.width || 50,
+      paddingLeft: style?.inProgress?.paddingLeft || 0,
+      paddingRight: style?.inProgress?.paddingRight || 0,
+      flex: style?.inProgress?.flex || 1,
+      alignSelf: style?.inProgress?.alignSelf || 'center',
+      opacity: style?.inProgress?.opacity || 0.5,
     };
-    delete style.underlayColor;
+  }
+
+  render() {
+    const { children, inProgress, style, ...otherProps } = this.props;
+
+    const resolvedStyle = this.resolveStyle();
 
     return (
       <TouchableOpacity
-        {...this.props}
-        style={style}
-        underlayColor={this.props.style.underlayColor}
-      />
+        {...otherProps}
+        style={resolvedStyle}
+        underlayColor={style?.underlayColor}
+        disabled={inProgress}
+      >
+        {inProgress && <Spinner />}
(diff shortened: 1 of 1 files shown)


components/Button.js
```diff
@@ -1,23 +1,50 @@
...
 
 class Button extends PureComponent {
-  render() {
-    // The underlayColor is not a valid RN style
-    // property, so we have to unset it here.
-    const style = {
-      ...this.props.style,
+  componentDidUpdate() {
+    LayoutAnimation.easeInEaseOut();
+  }
```
Definitely-Not-Vlad, line 10: I don't think this should be done on every single update, won't this produce a flicker?

components/Button.js
```diff
@@ -1,23 +1,50 @@
...
     return (
       <TouchableOpacity
-        {...this.props}
-        style={style}
-        underlayColor={this.props.style.underlayColor}
-      />
+        {...otherProps}
+        style={resolvedStyle}
+        underlayColor={style?.underlayColor}
+        disabled={inProgress}
```
Definitely-Not-Vlad, line 43: We can now no longer pass custom `disabled` prop, which I think is a loss of functionality. I also can't think of a way to meaningfully connect `inProgress` and custom `disabled`.

components/Button.js
```diff
@@ -26,6 +53,10 @@ Button.propTypes = {
   ...TouchableOpacity.propTypes,
 };
 
+Button.defaultProps = {
+  inProgress: undefined,
```
Definitely-Not-Vlad, line 57: `false` rather than undefined, just to make it a little bit cleaner.

Also spread `TouchableOpacity.defaultProps` in `Button.defaultProps`.

Definitely-Not-Vlad (changes requested):
Other than requesting changes, I generally think this is not something we should do with this component.

The developer should handle this by sending `Spinner` as a child component, as well as using `disabled` as they see fit.

components/Button.js
tomislav-arambasic: You're right

components/Button.js
```diff
@@ -1,23 +1,55 @@
...
+      ...style,
+      ...style?.inProgress,
+      borderRadius: style?.inProgress?.borderRadius || 50,
+      minWidth: style?.inProgress?.minWidth || 30,
+      width: style?.inProgress?.width || 50,
+      paddingLeft: style?.inProgress?.paddingLeft || 0,
+      paddingRight: style?.inProgress?.paddingRight || 0,
+      flex: style?.inProgress?.flex || 1,
+      alignSelf: style?.inProgress?.alignSelf || 'center',
+      opacity: style?.inProgress?.opacity || 0.5,
```
ltatarev, line 34: Personally not a fan of button changing style depending on `inProgress`. In my opinion it should look the same (width-wise) with spinner in center. 
I feel like these style resolving could be done a bit better, there are too many `?.` and fallbacks here for my taste - why not add `style`/`style.inProgress` to default props? In that case `[style, inProgress & style.inProgress]` would be all you need here 🤔

ltatarev (changes requested):

components/Button.js
```diff
@@ -1,23 +1,55 @@
...
-    // The underlayColor is not a valid RN style
-    // property, so we have to unset it here.
-    const style = {
-      ...this.props.style,
+  componentDidUpdate(prevProps) {
+    const { inProgress: prevInProgress } = prevProps;
+    const { inProgress } = this.props;
+
+    if (prevInProgress !== inProgress) {
+      LayoutAnimation.easeInEaseOut();
```
sstimac, line 13: This won't take effect for the first change right ? First transition from not loading to loading

components/Button.js
sstimac: Additionally, these styles should be inside the theme, not inline in defaultProps or here. But yeah, style shouldn't change here, only child content

components/Button.js
```diff
@@ -1,23 +1,55 @@
...
+      alignSelf: style?.inProgress?.alignSelf || 'center',
+      opacity: style?.inProgress?.opacity || 0.5,
     };
-    delete style.underlayColor;
+  }
+
+  render() {
+    const { children, inProgress, style, ...otherProps } = this.props;
+
+    const resolvedStyle = this.resolveStyle();
```
sstimac, line 41: you would need to memo this if you would go on with different styling, which we won't do

components/Button.js
```diff
@@ -1,23 +1,55 @@
...
+      paddingRight: style?.inProgress?.paddingRight || 0,
+      flex: style?.inProgress?.flex || 1,
+      alignSelf: style?.inProgress?.alignSelf || 'center',
+      opacity: style?.inProgress?.opacity || 0.5,
     };
-    delete style.underlayColor;
+  }
+
+  render() {
+    const { children, inProgress, style, ...otherProps } = this.props;
```
sstimac, line 39: inProgress is lacking context. Rename to loading

components/Button.js
sstimac: I'd go with loading || disabled. You never want a clickable element when something is loading so that's fine

sstimac (changes requested):

sstimac:
I like this improvement. I personally don't want to conditionally render this every time ( unless i have a custom loading spinner i want to use ). Just need to clean up a bit

tomislav-arambasic:
Closing, newly designed button animations are better and we should move it to be a part of our UI lib.

PR #669, opened 2022-01-03
Add react-native-device-info and isNotchedAndroid to UI

Definitely-Not-Vlad:
- added `isNotchedAndroid`
  - uses `hasNotch()` from `react-native-device-info` as well as `Platform` from `react-native`.
- added `notchedAndroid` namespace to `Device.select()`.
- used `notchedAndroid` and `StatusBar.currentHeight` in `theme.js` to properly resolve `StatusBar` height for notched Android devices
- removed `react-native-status-bar-height` package and replaced it's usage with just `StatusBar.currentHeight` from `react-native`


package.json
```diff
@@ -24,12 +23,13 @@
     "lodash": "4.17.21",
     "prop-types": "^15.7.2",
     "qs": "6.9.3",
+    "react-native-device-info": "~8.4.8",
```
sstimac, line 26: Let's zapeči here. This package should be safe in combination with the RN version, but let's be sure

sstimac (changes requested):
nice. Only one minor cr

package.json
Definitely-Not-Vlad: Were it any other dep, I would agree, but this one regularly receives new devices to and info about them (specifically we're interested in `hasNotch` list).

I've checked, they've not had an instance of patch versions causing any issues, they take semver seriously.

Unless you insist, I'd really like to keep the `~`.

sstimac (commented):
👍

sstimac (approved):

PR #672, opened 2022-01-13
Set specific versions of deps and devDeps

Definitely-Not-Vlad:
Using NPM's semver calculator I determined which versions were in the range we had set and then used those specific versions.


sstimac (approved):
undo the shoutem deps and we gucci

PR #694, opened 2022-03-02
Add multiple icons

Definitely-Not-Vlad:
Added all reusable generic icons being registered in Shoutem extensions:
- attach-media (image icon with + badge)
- clock
- download (down arrow inside a circle)
- garbage-can (delete)
- location-pins (two location pins)
- restore
- sleep (moon and stars)


tomislav-arambasic (approved):

sstimac (approved):
awesome. We will make this a convention for all new features that bring in the newly designed icons ! @tomislav-arambasic check if you can use some of this and avoid registration from extension / check if you are registering same icons possibly

PR #695, opened 2022-03-04
Feature/add animated loading

tomislav-arambasic:
- added new component, `AnimatedLoading`. It's a wrapper that hides children components & shows loading animation while `loading` prop is true
- added `lottie-react-native` dependency (fixed version)
- created `usePreviousValue` hook. We will use it to determine when some prop changed when compared to previous value, like we did in `componentDidUpdate`. This component probably shouldn't take care of this - component rendering `AnimatedLoading` should be sending proper loading states and not re-render `AnimatedLoading`, but it's a failsafe for when things don't work properly in parent component. This operations makes sure loading animation animates properly, at almost no-cost 
<br>

**Code implementation example:**
```
<AnimatedLoading
    loading={isPosting}
    style={style.loadingContainer}
>
  <TouchableOpacity
     onPress={handlePostPress}
      disabled={postButtonDisabled}
  >
     <Text>Post</Text>
  </TouchableOpacity>
</AnimatedLoading>
```

<br>

**Video presentation:**
https://user-images.githubusercontent.com/57353746/156716719-bb766045-4ee4-4386-937c-b6b9735997a4.mov


assets/index.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable global-require */
+export const animations = {
+  buttonLoading: require('./animations/buttonLoading.json'),
```
ltatarev, line 3: I don't see this animation being tied anywhere directly to a button. I know that is the intended use case, but overall this is "just" a loading animation and it could possibly be used with something else.
Would prefer if it was just called `loading`

ltatarev (approved):
Minor naming suggestion, leaving up to you.

assets/index.js
tomislav-arambasic: I've (shamelessly :D) stole your LoadingButton code and removed button logic. Guess who forgot to remove/rename this one :D

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+function AnimatedLoading({
+  children,
+  finishAnimationCallback,
+  loading,
+  style,
+}) {
+  const prevLoading = usePreviousValue(loading);
+  const animateProgress = useRef(new Animated.Value(0)).current;
+
+  useEffect(() => {
```
sstimac, line 21: do you really need this ? useEffect triggers only on the control prop change. Meaning that you could just check if the loading is true when it triggers. If it is, it means that it changed from false to true, which is the step you are monitoring. Like this : 

> useEffect(() => {
>   if (loading) {
>     startAnimation();
>   }
> }, [startAnimation, loading])

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+import { connectStyle } from '@shoutem/theme';
+import { View } from '@shoutem/ui';
+import { animations } from '../assets';
+import { usePreviousValue } from '../hooks';
+
+const AnimatedView = Animated.createAnimatedComponent(View);
+
+function AnimatedLoading({
+  children,
+  finishAnimationCallback,
```
sstimac, line 14: onAnimationFinished. But i'm not convinced you need this. Why is that btw ?

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+import LottieView from 'lottie-react-native';
+import PropTypes from 'prop-types';
+import { connectStyle } from '@shoutem/theme';
+import { View } from '@shoutem/ui';
+import { animations } from '../assets';
+import { usePreviousValue } from '../hooks';
+
+const AnimatedView = Animated.createAnimatedComponent(View);
+
+function AnimatedLoading({
```
sstimac, line 12: LoadingContainer or something. Name is not clear enough

sstimac, line 12: allow sending the animation to this component, with the default one being the one you're using right now

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+
+  const startAnimation = useCallback(() => {
+    Animated.timing(animateProgress, {
+      toValue: 1,
+      duration: 300,
+      useNativeDriver: false,
+    }).start(finishAnimation);
+  }, [animateProgress, finishAnimation]);
+
+  const finishAnimation = useCallback(() => {
```
sstimac, line 35: am i crazy, or these animated values aren't used for anything right now ? You are simply conditionally rendering the lottie animation, which doesn't require you to animate anything in particular

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+    });
+  }, [animateProgress, finishAnimationCallback]);
+
+  return (
+    <AnimatedView style={style.container}>
+      {!loading && children}
+      {loading && (
+        <LottieView
+          style={style.loading}
```
sstimac, line 53: i don't se this style added in theme

components/AnimatedLoading.js
```diff
@@ -0,0 +1,85 @@
...
+          source={animations.loading}
+          colorFilters={style.animationFilters}
```
sstimac, line 55: or this one

theme.js
```diff
@@ -2944,4 +2944,15 @@ export default (variables = defaultThemeVariables) => ({
       borderWidth: 1,
     },
   },
+
+  'shoutem.ui.AnimatedLoading': {
+    container: {
+      width: getSizeRelativeToReference(44, 375, window.width),
```
sstimac, line 2950: since this is a generic component, we don't want to have fixed container dimensions. The idea is to wrap anything under this, and it will work. What would happen here if you had a child that is larger than these dimensions ? use onLayout or it's equivalent to get children dimensions. Apply the same to the animation container, and then position and scale the animation in the same container

sstimac (changes requested):
A couple of things that need fixing here. Also, i want to have the Button implemented in this fashion :D We don't have to do the button for now

components/AnimatedLoading.js
tomislav-arambasic: I know, I've done it like that in first place. I've explained why I've implemented this in PR description :D 

> created usePreviousValue hook. We will use it to determine when some prop changed when compared to previous value, like we did in componentDidUpdate. This component probably shouldn't take care of this - component rendering AnimatedLoading should be sending proper loading states and not re-render AnimatedLoading, but it's a failsafe for when things don't work properly in parent component. This operations makes sure loading animation animates properly, at almost no-cost

theme.js
tomislav-arambasic: I've applied these changes and I'm strongly against it, we can discuss when you'll be able to.
This component will sometimes be used without children, like we do in social now.
When it is used without children, LoadingContainer flexes to parent, that can be full screen e.g., while children container was only 100 px height... (hard to explain like this :D)
The fact is, it can break layout and I'm strongly against defining View inside LoadingContainer only to adjust its dimensions properly.

On the other hand, code that I've had and you commented here is much more what we already do everywhere in Shoutem.
Those "unused" styles you mentioned - those are for users to override and set their desired dimensions for this container.
These default dimensions (minWidth and minHeight now) only assure  loading container to be enough width and height to show loading.
If user wants container to be specific dimensions - user will pass `style={{ container: {width: 300, height: 300} }}` to this component. Animation container will have loading animation centered and flexed.
If loading animation container is too big because it flexed, user will then pass style.loading - again, it'll be centered.

example:
```
style={{
              container: {
                width: 300,
                height: 300,
              },
              loading: {
                width: 100,
                height: 100,
              },
```

<br><br>
**Video:**

https://user-images.githubusercontent.com/57353746/156996062-3a46f8a8-a457-4033-90dd-cac80ddcc81b.mov

tomislav-arambasic: Ignore my previous comment.
With Slavko's help, I've applied suggested changes, component will now have dimensions per children layout dimensions.

assets/animations/loading.json
```diff
@@ -0,0 +1 @@
...r":{"a":0,"k":0,"ix":10},"p":{"a":0,"k":[114,72,0],"ix":2,"l":2},"a":{"a":0,"k":[114,72,0],"ix":1,"l":2},"s":{"a":1,"k":[{"i":{"x":[0.667,0.667,0.667],"y":[1,1,1]},"o":{"x":[0.333,0.333,0.333],"y":[0,0,0]},"t":28,"s":[100,100,100]},{"i":{"x":[0.667,0.667,0.667],"y":[1,1,1]},"o":{"x":[0.333,0.333,0.333],"y":[0,0,0]},"t":35,"s":[150,150,100]},{"t":42,"s":[100,100,100]}],"ix":6,"l":2}},"ao":0,"shapes":[{"ty":"gr","it":[{"ind":0,"ty":"sh","ix":1,"ks":{"a":0,"k":{"i":[[-6.627,0],[0,-6.627],[6.627,0],[0,6.627]],"o":[[6.627,0],[0,6.627],[-6.627,0],[0,-6.627]],"v":[[0,-12],[12,0],[0,12],[-12,0]],"c":true},"ix":2},"nm":"Path 1","mn":"ADBE Vector Shape - Group","hd":false},{"ty":"fl","c":{"a":0,"k":[0.136999990426,0.122000002394,0.125,1],"ix":4},"o":{"a":0,"k":100,"ix":5},"r":1,"bm":0,"nm":"Fill 1","mn":"ADBE Vector Graphic - Fill","hd":false},{"ty":"tr","p":{"a":0,"k":[114,72],"ix":2},"a":{"a":0,"k":[0,0],"ix":1},"s":{"a":0,"k":[100,100],"ix":3},"r":{"a":0,"k":0,"ix":6},"o":{"a":0,"k":100,"ix":7},"sk":{"a":0,"k":0,"ix":4},"sa":{"a":0,"k":0,"ix":5},"nm":"Transform"}],"nm":"Group 1","np":2,"cix":2,"bm":0,"ix":1,"mn":"ADBE Vector Group","hd":false}],"ip":0,"op":46,"st":0,"bm":0}],"markers":[]}
```
sstimac, line 1: rename this to dots or something more descriptive

components/ListView.js
```diff
@@ -165,11 +165,18 @@ class ListView extends PureComponent {
...
-    if (!ListEmptyComponent) {
+    // if list is empty and user haven't explicitly set ListEmptyComponent to null,
+    // use default ListEmptyComponent
+    if (ListEmptyComponent === undefined) {
       mappedProps.ListEmptyComponent = this.renderListEmptyComponent();
     }
 
+    // if list is empty and user explicitly set ListEmptyComponent prop to null,
+    // render nothing
+    if (ListEmptyComponent === null) {
```
sstimac, line 176: do you need this ?

components/LoadingContainer.js
```diff
@@ -0,0 +1,76 @@
...
+      height: dimensions.height * animationScale,
+    }),
+    [animationScale, dimensions],
+  );
+
+  return (
+    <View style={style.container}>
+      {!loading && <View onLayout={onLayoutChange}>{children}</View>}
+      {loading && (
+        <View style={[dimensions, style.container]}>
```
sstimac, line 39: set dimensions as second array member, to avoid faulty style override possibility

theme.js
```diff
@@ -2944,4 +2944,12 @@ export default (variables = defaultThemeVariables) => ({
       borderWidth: 1,
     },
   },
+
+  'shoutem.ui.LoadingContainer': {
+    container: {
+      backgroundColor: variables.paperColor,
```
sstimac, line 2950: let's remove this. No color should play nicely with all types of backgrounds

sstimac (approved):
Very nice. Just a few minor comments left

components/ListView.js
```diff
@@ -357,7 +359,7 @@ ListView.propTypes = {
   getSectionId: PropTypes.func,
   hasFeaturedItem: PropTypes.bool,
   keyExtractor: PropTypes.func,
-  ListEmptyComponent: PropTypes.node,
+  ListEmptyComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
```
Definitely-Not-Vlad, line 362: It can't be `PropTypes.node`? Just double checking, not sure myself.

components/LoadingContainer.js
```diff
@@ -0,0 +1,76 @@
...
+  animation,
+  children,
+  loading,
+  animationScale,
+  style,
+}) {
+  const [dimensions, setDimensions] = useState({});
+
+  function onLayoutChange(event) {
+    const { width, height } = event.nativeEvent.layout;
```
Definitely-Not-Vlad, line 18: Is this worth using `useCallback` on with `[event.nativeEvent.layout.width, event.nativeEvent.layout.height]`?

Trying to learn, more a question than a suggestion.

Definitely-Not-Vlad (approved):

components/ListView.js
tomislav-arambasic: I can't remember last time it was node.
In Shoutem, it's usually a function or object - function for "plain" component, object if component is wrapped in connectStyle (if I remember right, last time I checked)

components/LoadingContainer.js
tomislav-arambasic: event is received through function props and it doesn't belong to dependency array - it can't be there - breaks the app with "Can't find variable: event".
Only constants present in component can be depended on

PR #697, opened 2022-03-06
Feature/remove vector icons

Definitely-Not-Vlad:
- remove large dependency `react-native-vector-icons` that was being used for one easily replaceable icon


ltatarev (approved):
MVP 🥇 🥇 🥇 🥇 🥇

sstimac (approved):
Awesome. Please check the new icons if they are being coloured properly when changing theme style. multiple path SVGs were known to cause some issues with this

PR #708, opened 2022-04-03
Fix cyclic dependency in keyboard helper, cleanup index

Definitely-Not-Vlad:
`keyboard.js` was importing something from `theme.js`, while also being imported into `theme.js`. No bueno.

### DIFF
--- a/const.js
+++ b/const.js
@@ -1,3 +1,6 @@
+const { Platform, StatusBar } = require('react-native');
+const { Device } = require('./helpers');
+
 const IPHONE_X_HOME_INDICATOR_PADDING = 34;
 const IPHONE_X_LONG_SIDE = 812;
 const IPHONE_X_NOTCH_PADDING = 30;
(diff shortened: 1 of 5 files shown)


ltatarev (changes requested):
Requesting change only for the Status Bar thingy because I think that should be fixed, other than that LGTM

const.js
```diff
@@ -10,6 +13,15 @@ const IPHONE_12_MAX_LONG_SIDE = 926;
 
 const NAVIGATION_HEADER_HEIGHT = 64;
 
+const STATUS_BAR_OFFSET =
+  Platform.OS === 'android' ? -StatusBar.currentConfig : 0;
+const NAVIGATION_BAR_HEIGHT = Device.select({
+  iPhoneX: NAVIGATION_HEADER_HEIGHT + IPHONE_X_NOTCH_PADDING,
+  iPhoneXR: NAVIGATION_HEADER_HEIGHT + IPHONE_XR_NOTCH_PADDING,
+  notchedAndroid: NAVIGATION_HEADER_HEIGHT + StatusBar.currentHeight,
```
tomislav-arambasic, line 21: Not sure if notchedAndroid is resolved properly, I think it applies StatusBar height out of the box?

tomislav-arambasic (changes requested):
Requesting changes only to get an answer on a single comment.
Thanks again for doing this :lovem:

sstimac (approved):

ltatarev (approved):

PR #709, opened 2022-04-06
Feature/apply prop types conventions

Definitely-Not-Vlad:
- moved all `static propTypes` and `static defaultProps` to `ComponentName.propTypes` and `ComponentName.defaultProps` after class component definition
- set defaults where they were provided later in props destructuring / in constructor
- set `style` as `isRequired` in any component using `connectStyle` as it provides a default of `{}` if not the style for the component, meaning it won't ever be `undefined`

Non-prop-type related changes:
- `{ Icon }` -> `Icon` import correction in `ImagePreview`
- import ordering auto-format in `InlineDropDownMenu`
- unused lodash import removal in `InlineDropDownMenuItem`,

### DIFF
--- a/components/ActionSheet/ActionSheet.js
+++ b/components/ActionSheet/ActionSheet.js
@@ -33,14 +33,6 @@ class ActionSheet extends PureComponent {
     return null;
   }
 
-  static propTypes = {
-    style: PropTypes.any,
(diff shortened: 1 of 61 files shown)


sstimac (approved):
Praise the Vlad

tomislav-arambasic (approved):

PR #710, opened 2022-04-06
Feature/apply prop and state destructuring

Definitely-Not-Vlad:
Follow up PR for prop type and default prop convention.

- destructure `this.props` and `this.state` wherever necessary
- replace weird `const { props } = this;` usage with normal prop destructuring
- removed all possible uses of `delete`, replaced them with `const { deletedProperty, ...otherProperties } = object;`


components/DropDownMenu/DropDownModal.js
```diff
@@ -66,9 +66,9 @@ class DropDownModal extends PureComponent {
   }
 
   emitOnOptionSelectedEvent(option) {
-    if (this.props.onOptionSelected) {
-      this.props.onOptionSelected(option);
-    }
+    const { onOptionSelected } = this.props;
+
+    onOptionSelected && onOptionSelected(option);
```
tomislav-arambasic, line 71: I'm not sure if we have a convention on this, but we should.

I'm fine with both,`if` or ternary conditional callbacks, but we should agree on one of those.

I think I'd prefer `if` more, kinda easier to read

components/ImageGallery/ImageGalleryBase.js
```diff
@@ -29,11 +29,15 @@ export class ImageGalleryBase extends PureComponent {
 
     autoBindReact(this);
 
+    const { selectedIndex } = props;
+
     this.state = {
-      selectedIndex: this.props.selectedIndex || 0,
+      selectedIndex,
```
tomislav-arambasic, line 35: You sure we don't need fallback `0` here?
If we do, please add it to `defaultProps`

components/ScrollView/ScrollView.js
```diff
@@ -77,19 +77,17 @@ class ScrollView extends PureComponent {
...
+    const { contentContainerStyle, ...otherStyle } = style;
 
     return (
       <Animated.ScrollView
         ref={this.setWrappedInstance}
         contentContainerStyle={this.addIphoneXPadding(contentContainerStyle)}
-        {...animationDriver.scrollViewProps}
-        {..._.omit(props, 'onScroll')}
+        {...scrollViewProps}
+        {..._.omit(otherProps, 'onScroll')}
```
tomislav-arambasic, line 89: Bruh, we need to memoize this asap (ik it wasn't part of this PR, just sayin')

components/View.js
```diff
@@ -7,25 +7,23 @@ import { LinearGradient } from './LinearGradient';
...
 class View extends PureComponent {
   render() {
-    const style = { ...this.props.style };
-    let gradient = null;
+    const { children, style, ...otherProps } = this.props;
+    const { backgroundGradient, ...viewStyle } = style;
 
-    if (style.backgroundGradient) {
+    let gradient = null;
+    if (backgroundGradient) {
```
tomislav-arambasic, line 14: Empty line above please

tomislav-arambasic (changes requested):
Thanks for doing this.
Requesting changes only because of fallback 0 value, I'm not sure if not having it could break things

components/ImageGallery/ImageGalleryBase.js
Definitely-Not-Vlad: Already done in previous PR in chain where defaultProps are defined.

components/DropDownMenu/DropDownModal.js
Definitely-Not-Vlad: This is not ternary, this is just how `&&` works, if both are true, it returns the right hand side, so you can expect it to execute the function passed to it.

I prefer this, obviously, but we can put it to vote in a dev-weekly, I'll add it to the agenda as a tiny point.

Until then, I'll use `if (x) { x() }`

sstimac: safest would be if (_.isFunction(callback)), but we could also set this to requiredProp and resolve the issue that way

sstimac (approved):
very nice. thanks for doing this

tomislav-arambasic (approved):

PR #721, opened 2022-05-24
Fix style resolution

tomislav-arambasic:
We gotta have fallback for `contentContainerStyle`, otherwise ScrollViews will crash the app

### DIFF
--- a/components/ScrollView/ScrollView.js
+++ b/components/ScrollView/ScrollView.js
@@ -79,7 +79,7 @@ class ScrollView extends PureComponent {
   render() {
     const { style, ...otherProps } = this.props;
     const { scrollViewProps } = this.animationDriver;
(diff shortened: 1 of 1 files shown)


components/ScrollView/ScrollView.js
```diff
@@ -79,7 +79,7 @@ class ScrollView extends PureComponent {
   render() {
     const { style, ...otherProps } = this.props;
     const { scrollViewProps } = this.animationDriver;
-    const { contentContainerStyle, ...otherStyle } = style;
+    const { contentContainerStyle = {}, ...otherStyle } = style;
```
sstimac, line 82: can't we just add this style inside the theme directly ?

sstimac (changes requested):

Definitely-Not-Vlad (approved):

sstimac (approved):

PR #742, opened 2022-07-27
Fix circular dependencies

ltatarev:
Feature fixes 2 circular deps:
- `const.js > helpers/index.js > helpers/device-selector.js`
- `const.js > helpers/index.js > helpers/keyboard.js`
which resulted in `calculateKeyboardOffset` returning `NaN`


Definitely-Not-Vlad (approved):
LGTM, probably something I broke to begin with, so thank you for cleaning up my mess

PR #743, opened 2022-07-28
Update exports 🤦‍♀️

ltatarev:
Forgot to export all consts moved to device-selector yesterday.. I really hope we can rewrite and update this service entirely sometime soon..


Definitely-Not-Vlad (approved):
I agree about the rewrite, this is not only ugly, but also annoying to use.

PR #745, opened 2022-08-03
Add `dimensionRelativeToIphoneHeight` helper

ltatarev:
Since `dimensionRelativeToIphone` takes into consideration only width, I added the same helper function for height, because I think this is something we should support as well.
Not sure if naming of both functions could be improved though.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,12 @@ export {
   calculateLineHeight,
   defaultThemeVariables,
   dimensionRelativeToIphone,
-  dimensionRelativeToIphoneHeight,
   default as getTheme,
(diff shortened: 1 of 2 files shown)


Definitely-Not-Vlad (commented):
Ideally we'd deprecate `dimensionRelativeToIphone` and tell people to replace it with `dimensionRelativeToIphoneWidth`, or whatever better name we think of.

I suggest something like `responsiveHeight(200)` and `responsiveWidth(200)`, because that's what it is basically.

Will postpone approval until this is implemented.

Definitely-Not-Vlad (approved):

PR #746, opened 2022-08-10
Add DropShadowView component

ltatarev:
Feature adds `DropShadowView` component using `react-native-drop-shadow`.
Slightly adjusted `View` component and gradient resolution.


Definitely-Not-Vlad (approved):
Approved but please hold off on merging until we're ready to add this to Disclose.

PR #748, opened 2022-08-17
Fix lottie dependencies and bump versions

sstimac:


toroz (approved):

Definitely-Not-Vlad (changes requested):
Alphabetic order please.

Definitely-Not-Vlad (approved):

PR #749, opened 2022-08-17
Release/5.3.1 -> master

sstimac:


toroz (approved):

components/Image.js
```diff
@@ -82,7 +83,7 @@ class Image extends PureComponent {
 }
 
 Image.propTypes = {
-  ...RNImage.propTypes,
+  ...ImagePropTypes.propTypes,
```
Definitely-Not-Vlad, line 86: I know this isn't from your previous PR, but this is an error, it's just supposed to be `...ImagePropTypes,`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "5.3.0",
+  "version": "5.3.1",
```
Definitely-Not-Vlad, line 3: With the other changes in Develop, this is a minor version, `5.4.0`, because we added `responsiveWidth` and `responsiveHeight` helpers, as well as added deprecation warning for `dimensionRelativeToIphone` which is to be removed with 6.0.0 eventually, being replaced with `responsiveWidth`.

Definitely-Not-Vlad (changes requested):

Definitely-Not-Vlad (approved):

PR #753, opened 2022-09-30
Remove examples

ltatarev:
Feature removes `examples` dir, updates readme, and removes remaining references to examples.
Additionally, removed `nativeDependencies` export from `const`, since the variable doesn't exist (noticed due to lint error).


Definitely-Not-Vlad (approved):
How come we're removing this? Debloat or?

ltatarev:
> How come we're removing this? Debloat or?

We agreed to do this a while ago on a weekly meeting? :D 
If I remember correctly, the main reason was that they aren't maintained, and that it would be nicer to go in a storybook direction

Definitely-Not-Vlad:
Ah, I figured we'd remove them after adding storybook but this works I guess.

sstimac (approved):
bello, bello indescrivibile

PR #759, opened 2022-10-13
Update ImageGallery & Lightbox

ltatarev:
Feature removes usage of forked repos for image components.

`ImageGallery` was rewritten in hooks, and platform specific code was removed.
Needed image gestures are added via ReactNativeZoomableView which behaves the same way as photo view/transformable image + the library is well maintained.

`Lightbox` library was replaced with lightbox v2 which offers the same feature as previously used fork.

I've tested changes on both platform devices and from what I could tell both components behave the same way as before.


components/ImageGallery.js
```diff
@@ -0,0 +1,152 @@
...
+  renderImageOverlay,
+  renderPlaceholder,
+  selectedIndex,
+  showNextPage,
+  style,
+}) {
+  const [currentIndex, setCurrentIndex] = useState(selectedIndex);
+  const [mode, setMode] = useState(IMAGE_GALLERY_MODE);
+
+  function onImageTap(newMode) {
```
sstimac, line 29: rename with handle prefix and use useCallback. apply elsewhere

components/ImageGallery.js
```diff
@@ -0,0 +1,152 @@
...
+    return (
+      <ReactNativeZoomableView
+        maxZoom={1.5}
+        minZoom={0.5}
+        zoomStep={0.5}
+        initialZoom={1}
+        bindToBorders
+      >
+        <Pressable onPress={onImageTap}>
+          <Image style={[imageStyle, style.image]} {...resolvedImageProps} />
```
sstimac, line 62: let's memo the style

sstimac (approved):

components/ImageGallery.js
ltatarev: I don't see an easy way to do this since `imageStyle` is a prop from `renderPage` and it changes for each page

PR #760, opened 2022-10-19
Update camera and gallery svg

ltatarev:
Both svgs had hardcoded `fill="black"` so it wasn't possible to overwrite their color through style.


sstimac (approved):
nice catch

PR #763, opened 2022-10-24
Fix gallery crash

ltatarev:
Feature fixes `ImageGallery` crash caused by `transformImageProps` expecting `style` as a prop.

Not sure if this should be handled here or on extensions side, since shoutem.application defines [image prop transformer](https://github.com/shoutem/extensions/blob/850c0854f5008072c3d4d00d36f7d179dc65f9b7/shoutem.application/app/services/resizeImageSource.js#L22) that causes the crash. 
It used to work properly since old component had defined `style: { flex: 1 }` here.

### DIFF
--- a/components/ImageGallery.js
+++ b/components/ImageGallery.js
@@ -85,6 +85,7 @@ function ImageGallery({
     const transformImageProps = Image.getPropsTransformer();
     const imageProps = {
       source: { uri: image },
+      style: { ...style.image },
     };
     const transformedImageProps = _.isFunction(transformImageProps)
       ? transformImageProps(imageProps)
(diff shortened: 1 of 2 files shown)


toroz (approved):

components/ImageGallery.js
```diff
@@ -85,6 +85,7 @@ function ImageGallery({
     const transformImageProps = Image.getPropsTransformer();
     const imageProps = {
       source: { uri: image },
+      style: { ...style.image },
```
sstimac, line 88: I'd say this is okay, just remove the spread as it doesn't feel necessary

sstimac (approved):

PR #764, opened 2022-11-10
Feature/app level toasts

sstimac:
- Our take on the https://github.com/calintamas/react-native-toast-message implementation
- Exposed Shoutem styled toasts that cover most of the scenarios ( success, action with buttons, info, error )
- "Native" api still accessible if you don't want shoutem styled toasts, and just want to use something completely custom


components/ToastMessage/components/InfoToast.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { connectStyle } from '@shoutem/theme';
+import BaseToast from './BaseToast';
+
+function InfoToast({ style, props, isVisible }) {
+  return (
+    <BaseToast customToastStyle={style} isVisible={isVisible} {...props} />
```
Definitely-Not-Vlad, line 7:
```suggestion
function InfoToast({ style, isVisible, ...otherProps }) {
  return (
    <BaseToast customToastStyle={style} isVisible={isVisible} {...otherProps} />
```

Just a suggestion, I've seen this notation elsewhere and it makes more sense to me because `style` and `isVisible` are also props, so having them next to a prop called `props` seems unusual.

components/ToastMessage/components/SuccessToast.js
```diff
@@ -0,0 +1,19 @@
...
+import { images } from '../assets';
+import BaseToast from './BaseToast';
+
+function SuccessToast({ style, props, isVisible }) {
+  return (
+    <BaseToast
+      customToastStyle={style}
+      iconSource={images.success}
+      isVisible={isVisible}
+      {...props}
```
Definitely-Not-Vlad, line 12:
```suggestion
function SuccessToast({ style, isVisible, ...otherProps }) {
  return (
    <BaseToast
      customToastStyle={style}
      iconSource={images.success}
      isVisible={isVisible}
      {...otherProps}
```

Just a suggestion, I've seen this notation elsewhere and it makes more sense to me because `style` and `isVisible` are also props, so having them next to a prop called `props` seems unusual.

components/ToastMessage/components/ErrorToast.js
```diff
@@ -0,0 +1,19 @@
...
+import { images } from '../assets';
+import BaseToast from './BaseToast';
+
+function ErrorToast({ style, props, isVisible }) {
+  return (
+    <BaseToast
+      customToastStyle={style}
+      isVisible={isVisible}
+      iconSource={images.error}
+      {...props}
```
Definitely-Not-Vlad, line 12:
```suggestion
function ErrorToast({ style, isVisible, ...otherProps }) {
  return (
    <BaseToast
      customToastStyle={style}
      isVisible={isVisible}
      iconSource={images.error}
      {...otherProps}
```

Just a suggestion, I've seen this notation elsewhere and it makes more sense to me because `style` and `isVisible` are also props, so having them next to a prop called `props` seems unusual.

components/ToastMessage/components/ActionToast.js
```diff
@@ -0,0 +1,19 @@
...
+import { images } from '../assets';
+import BaseToast from './BaseToast';
+
+function ActionToast({ style, isVisible, props }) {
+  return (
+    <BaseToast
+      customToastStyle={style}
+      imageSource={images.action}
+      isVisible={isVisible}
+      {...props}
```
Definitely-Not-Vlad, line 12:
```suggestion
function ActionToast({ style, isVisible, ...otherProps }) {
  return (
    <BaseToast
      customToastStyle={style}
      imageSource={images.action}
      isVisible={isVisible}
      {...otherProps}
```

Just a suggestion, I've seen this notation elsewhere and it makes more sense to me because `style` and `isVisible` are also props, so having them next to a prop called `props` seems unusual.

components/ToastMessage/assets/index.js
```diff
@@ -0,0 +1,9 @@
+const action = require('./action.png');
+const success = require('./success.png');
+const error = require('./error.png');
+
+export const images = {
+  action,
+  success,
+  error,
+};
```
Definitely-Not-Vlad, line 9: Is there a reason we're using PNGs instead of SVG icons?

Definitely-Not-Vlad (commented):
Not requesting any changes, but can't approve this either because I don't really understand `toastMessage.js` without docs, so I'll wait for that.

Take a peek at the change suggestions in the comments in this review, let me know what you think.

components/ToastMessage/components/InfoToast.js
sstimac: props here is actually a named prop, coming from the toast component. Can't do much here other than renaming for more clarity

components/ToastMessage/assets/index.js
sstimac: i can change. Not sure why i exported pngs from figma

components/ToastMessage/components/BaseToast.js
```diff
@@ -0,0 +1,123 @@
+import React, { useMemo } from 'react';
+import _ from 'lodash';
+import PropTypes from 'prop-types';
+import { connectStyle } from '@shoutem/theme';
+import { Icon, Image, Text, Touchable, View } from '@shoutem/ui';
```
ltatarev, line 5: Imports directly from components :D

components/ToastMessage/toastMessage.js
```diff
@@ -0,0 +1,54 @@
...
+
+export default {
+  Provider,
+  showAction,
+  showError,
+  showSuccess,
+  showInfo,
+  show: Toast.show,
+  hide: Toast.hide,
+  defaultConfig: config,
```
ltatarev, line 53: Curious why was this exported, I don't think it's neccessary? Is there an option to pass custom config to Toast provider, or is that not intended?

components/ToastMessage/assets/index.js
ltatarev: +1 for changing to svgs if possible

ltatarev (approved):
Really happy this is finally getting added. `BaseToast` looks `really` good, big fan of all the customizability options.

components/ToastMessage/toastMessage.js
sstimac: I exported so that you can build on top of your config. So you want to retain all of our toasts but include more custom ones. Provider is also exported so you can manipulate it's props with custom config

Definitely-Not-Vlad (approved):
Great docs! LGTM

PR #793, opened 2023-12-13
Feature/adjust with notch padding calculation

tomislav-arambasic:
## Summary
Our `with-notch-padding` styleName was calculating padding for home indicator (not notch), and it was working only for selected iPhone models - iPhone X and XR. If user had newer devices, bottom padding wouldn't apply.

iPhone identifiers can be found here:
https://www.theiphonewiki.com/wiki/Models#iPhone

- Add new calculation for bottom padding, based on iPhone devices with home indicator. Devices with home indicator calculation is based on iPhone ID, a string containing Apple's internal version for each model, e.g. `iPhone16,2`
  - Home indicator is present on all iPhone devices with ID 10,3 or greater, except:
    - iPhone 8 and iPhone 8 plus newer generations
    - 2nd and 3rd gen iPhoneSE
- Created new styleName for `View`, `Screen` and `ListView`, with more appropriate name - `with-home-indicator-padding`, which should be used in place of `with-notch-padding`
- Adjusted `with-notch-padding` to use new calculation, while keeping backwards compatibility

### Screenshots

<table>
<tr>
<td>with-home-indicator-padding</td>
<td>NEW with-notch-padding</td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

### DIFF
--- a/helpers/device-selector.js
+++ b/helpers/device-selector.js
@@ -1,5 +1,5 @@
 import { Dimensions, Platform, StatusBar } from 'react-native';
-import { hasNotch } from 'react-native-device-info';
+import DeviceInfo from 'react-native-device-info';
 import _ from 'lodash';
 
 const { OS, isPad, isTVOS } = Platform;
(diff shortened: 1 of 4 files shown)


Definitely-Not-Vlad (approved):
LGTM. This should bump the minor version of UI before release as it is a new feature.

helpers/device-selector.js
```diff
@@ -71,6 +71,51 @@ export const NAVIGATION_BAR_HEIGHT = select({
...
+  // If version is present, e.g. in iPhone 9,1, it'll always return array of length 3.
+  if (!versions && versions.length < 3) {
+    return 0;
+  }
+
+  // Extracted numbers as strings
+  const majorVersion = versions[1];
+  const minorVersion = versions[2];
+
+  const decimalNumber = parseFloat(`${majorVersion}.${minorVersion}`);
```
sstimac, line 95: you can call this numericVersion or resolvedVersion

sstimac (approved):

PR #798, opened 2023-12-19
Feature/seext 12720 wall of blank space after video attachments from rte

tomislav-arambasic:
## Summary
Shoutem's RTE component is wrapping video attachment's `iframe` element inside `figure` element.
That `figure` element holds the style that breaks the UI - adds lots of blank space under video attachment.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

`figure` element can't be "turned off" by RTE component AND it also doesn't work if that element is removed with regex, so that's not an option.
We have to remove it on client side, prior to rendering, by manipulating the source html.

- Add `domVisitors` to `SimpleHtml` component
  - Define default, Shoutem specific `onElement` domVisitor, while still allowing component users to override `domVisitors`
  - `onElement` and `removeShoutemRteVideoAttachmentWrapper` are exported to give component users more control over defining `domVisitors`

## Side changes
- Fix Vimeo video attachments breaking UI - these video embeds wouldn't respect `SimpleHtml` right padding
  - Use `react-native-vimeo-iframe` to render Vimeo embeds as a part of iframes inside our `SimpleHtml`
  - Set the height for Vimeo embeds just enough to support vertical videos & render all Vimeo player controls.

### Screenshots
<table>
<tr>
<td colSpan="3">iOS</td>
</tr>
<tr>
<td></td>
<td></td>
<td></td>
</tr>

<tr>
<td colSpan="3">Android</td>
</tr>
<tr>
<td></td>
<td></td>
<td></td>
</tr>
</table>

### DIFF
--- a/theme.js
+++ b/theme.js
@@ -2327,6 +2327,10 @@ export default () => {
         height: responsiveWidth(194),
         paddingBottom: resolveVariable('smallGutter'),
       },
+      // Just enough to display portrait mode Vimeo videos and all Vimeo player controls
+      vimeoVideo: {
+        height: responsiveWidth(220),
+      },
       fallback: {
         width: window.width,
         height: 40,
--- a/theme.js
+++ b/theme.js
@@ -2329,7 +2329,7 @@ export default () => {
       },
       // Just enough to display portrait mode Vimeo videos and all Vimeo player controls
       vimeoVideo: {
-        height: responsiveWidth(220),
+        height: responsiveHeight(220),
       },
       fallback: {
         width: window.width,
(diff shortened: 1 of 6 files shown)


theme.js
```diff
@@ -2327,6 +2327,10 @@ export default () => {
         height: responsiveWidth(194),
         paddingBottom: resolveVariable('smallGutter'),
       },
+      // Just enough to display portrait mode Vimeo videos and all Vimeo player controls
+      vimeoVideo: {
+        height: responsiveWidth(220),
```
Definitely-Not-Vlad, line 2332:
```suggestion
        height: responsiveHeight(220),
```
Are we sure we want to use `responsiveWidth` (note **_width_**) for `height`?

Added replacement with `responsiveHeight` instead, commit it if necessary, if not, ignore this comment.

Definitely-Not-Vlad (approved):
LGTM, but please check the comment I added and commit the suggestion if necessary.

sstimac (approved):

Issue #478, opened 2020-05-01
StatusBar style fixing


merkdev:
Hello there,

There is statusBar options in navigationBar component. But barStyle for statusBar is wrong written.

https://github.com/shoutem/ui/blob/5cc29e862a5c1da540b8f2a660b8f91e06abc74b/components/NavigationBar/NavigationBar.js#L45

im guessing; "default" should be replaced with 'dark-content'

Thanks!

abdulbasithqb:
Any update on this issue ?

PR #806, opened 2023-12-26
fix: statusBar style fixing

abdulbasithqb:
Enhance `chooseBarStyle` function in React Native component library to dynamically set the navigation bar style to `'light-content'` when the background color is light, ensuring improved visibility and maintaining design consistency across diverse backgrounds."

fixes: #478

### DIFF
--- a/components/NavigationBar/NavigationBar.js
+++ b/components/NavigationBar/NavigationBar.js
@@ -22,7 +22,7 @@ function getBackgroundColor(style) {
 class NavigationBar extends PureComponent {
   setStatusBarStyle(backgroundColor) {
     function chooseBarStyle(bgColor) {
-      return color(bgColor).isDark() ? 'light-content' : 'default';
+      return color(bgColor).isDark() ? 'light-content' : 'dark-content';
     }
 
     function setStyle(bgColor) {


tomislav-arambasic:
@abdulbasithqb Thanks for creating a PR for this.

@sstimac @Definitely-Not-Vlad 
This looks like a right change to me.
It doesn't change anything for iOS, but does on Androids.
`default` style for iOS is `dark` and `light` for Android, meaning that, currently, we never switch to `dark` StatusBar style on Androids. 

I don't see why we shouldn't merge this in, unless there was a specific reason behind implementing `default` style that I'm not aware of?
I believe all Androids have solid Status Bar background, so the `style` would be just a matter of personal preference, no UI breaking changes no matter of `style` chosen.

tomislav-arambasic (approved):

tomislav-arambasic:
@abdulbasithqb 
We can merge your changes. However, I need you to confirm that changes in this PR are fixing it for you.

I don't see how you get Status Bar changed on Android because of if-else condition.
On Androids, it never enters else condition on line 35 - never sets resolved bar style.

If above is true, please adjust your PR and verify everything works as expected.

PR #807, opened 2024-02-20
Hotfix/6.4.1 -> master

sstimac:
fix RN FlatList method mapping

### DIFF
--- a/components/ListView.js
+++ b/components/ListView.js
@@ -415,10 +415,31 @@ StyledListView.prototype.scrollTo = function scrollTo(coordinates) {
   }
 };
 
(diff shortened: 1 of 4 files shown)


sstimac:
> LGTM, but is it truly a patch change? We are switching from animation to config, I'm not sure what the difference is or if it's a breaking change or just a fix.
> 
> If it's just a fix, then we may as well consider it a patch.

It won't work by sending the boolean param. It needs to be config to work with the RN FlatList. So it's actually a fix :)

tomislav-arambasic (approved):

PR #808, opened 2024-02-20
Hotfix/6.4.1 -> develop

sstimac:
mind the false diff ....


Definitely-Not-Vlad (approved):
LGTM, but is it truly a patch change? We are switching from `animation` to `config`, I'm not sure what the difference is or if it's a breaking change or just a fix.

If it's just a fix, then we may as well consider it a patch.

PR #832, opened 2024-07-05
Add wheelchair icon

Definitely-Not-Vlad:


sstimac (approved):
Please hotfix this if it's going out sooner than the audio release. I think it is. Coordinate versions with @tomislav-arambasic

PR #844, opened 2024-08-28
Feature/support action sheet styling

tomislav-arambasic:
### Change default color of action sheet options to iOS kind of blue, when Platform is iOS or Web. This is how it looks natively.
Screenshot 2024-08-28 at 11 02 13

### Add option to override default styling, from inside implementations. Web is using this to get desired action sheet option text color, to match the one in simulator.
Screenshot 2024-08-28 at 10 52 19


sstimac (approved):
I'm okay with extending the style for the options via props, but i'm not sure about styling the component separately per platform.

PR #847, opened 2024-09-12
Fix/toast message height

tomislav-arambasic:
Removed fixed height for text container inside Toast message.
Toast will now use auto height, depending on the length of the text. Everything is centred as expected.

I thought about adding `minHeight`, setting it to the value toast had before, but I really don't think it's necessary - it even looks better when container is fit to only 2 lines of text (title & message).

If you think old UI looks better for shorter toast messages, let me know, I'll add minHeight. Check last 2 screenshots and their description, for comparison.

Long text (current state)
Screenshot 2024-09-12 at 20 30 48

Short text (current state)
Screenshot 2024-09-12 at 20 35 17

Short text with minHeight 48 (old state, UI)
Screenshot 2024-09-12 at 20 35 30

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@shoutem/ui",
-  "version": "7.1.0-beta.17",
+  "version": "7.1.0-beta.18",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


sstimac (approved):
LGBTQ+

PR #848, opened 2024-09-16
fix web fonts resolution

tomislav-arambasic:
We've switched from webpack to esbuild. Esbuild works differently with fonts and these changes are accommodating that difference.


sstimac (approved):
LGTM

PR #861, opened 2024-11-29
Feature/filters icon and searchinput adjustments

tomislav-arambasic:
- Added filters icon
- Adjusted SearchInput component
  - Add internal state, so that close button rendering logic can be resolved within the component itself. No need for having state outside component this way...
  - Rewrite to functional component


components/SearchField.js
```diff
@@ -23,59 +24,50 @@ ClearButton.propTypes = {
...
+    >
+      <Icon name="search" style={style.searchIcon} />
+      <TextInput
+        {...otherProps}
+        autoCapitalize="none"
+        autoCorrect={false}
+        onChangeText={setText}
+        placeholder={placeholder}
+        style={style.input}
+        value={text}
```
sstimac, line 53: move this  and auto props above otherProps spread

sstimac (approved):

PR #864, opened 2024-12-16
Feature/refactor size resolutions

sstimac:
- remove NavigationBar component ( deprecated from before ), and the set of matching styles
- rework size resolutions for Notched devices, devices with home padding, etc.. ( we are now device agnostic)
- adjust components referencing the old device size resolution


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shoutem/ui",
-  "version": "7.2.2",
+  "version": "7.3.0-rc.0",
```
tomislav-arambasic, line 3: v8

tomislav-arambasic (approved):
Please update version to v8, mainly because of breaking changes, there could be someone using those.

Otherwise, **magnificent** work 👏

PR #867, opened 2024-12-19
Feature/progress rings

tomislav-arambasic:
## Summary
- Add `ProgressRings` component for rendering filling ring UI representation of given percentage
  - There are 2 concepts available: `ProgressRing` and `AnimatedProgressRing`
  - If single, definite color is given as prop to ProgressRings `rings`, we'll render `ProgressRing` component colored as specified
  - If array of `progressColors` is given as prop, then `AnimatedProgressRing` is used as main component and it interpolates between given `progressColor`s based on given percentage value
  - We could potentially extend later component with filling animation, but I couldn't figure it out now, I'm also out of time. I've probably had issues with re-rendering in my implementation
- Add `useColorInterpolation` hook which resolves color based on given colors and values. For progress rings case, this is used to get the same color for rings & outside rings component, but in general, it can be used in various scenarios

<table>
<tr>
<td></td>
<td></td>
<td></td>
<td></td>
</tr>
</table>

https://github.com/user-attachments/assets/1cb23f5a-4878-4a49-a157-d683340346a8

### DIFF
--- a/components/ProgressRings/ProgressRing.js
+++ b/components/ProgressRings/ProgressRing.js
@@ -1,0 +1,85 @@
+/* eslint-disable react/require-default-props */
+import React from 'react';
+import { Path } from 'react-native-svg';
+import PropTypes from 'prop-types';
+import { changeColorAlpha } from '@shoutem/theme';
+import { PROGRESS_RING_DEFAULT_PROPS } from './const';
+import { circlePath } from './svgCalculations';
+
+const ProgressRing = ({
+  index,
+  width = PROGRESS_RING_DEFAULT_PROPS.width,
+  backgroundWidth = PROGRESS_RING_DEFAULT_PROPS.backgroundWidth,
+  progressLineCap = PROGRESS_RING_DEFAULT_PROPS.progressLineCap,
+  backgroundLineCap = PROGRESS_RING_DEFAULT_PROPS.backgroundLineCap,
+  percentage = PROGRESS_RING_DEFAULT_PROPS.percentage,
+  hasBackgroundColor = PROGRESS_RING_DEFAULT_PROPS.hasBackgroundColor,
+  arcSweepAngle = PROGRESS_RING_DEFAULT_PROPS.arcSweepAngle,
+  size = PROGRESS_RING_DEFAULT_PROPS.size,
+  color = PROGRESS_RING_DEFAULT_PROPS.color,
+}) => {
+  const halfSize = size / 2;
+
+  const maxWidthCircle = backgroundWidth
+    ? Math.max(width, backgroundWidth)
+    : width;
+
+  const radius = halfSize - maxWidthCircle / 2 - index * maxWidthCircle;
+
+  const currentFillAngle =
+    (arcSweepAngle * Math.min(100, Math.max(0, percentage))) / 100;
+
+  const backgroundPath = circlePath(
+    halfSize,
+    halfSize,
+    radius,
+    currentFillAngle,
+    arcSweepAngle,
+  );
+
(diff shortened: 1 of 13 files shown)


hooks/useColorInterpolation.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const resolveInterpolationRange = colors => {
+  return {
+    inputRange: resolveInputRange(colors),
+    outputRange: resolveOutputRange(colors),
+  };
+};
+
+export const useColorInterpolation = (colors, percentage) => {
```
sstimac, line 32: Based on `GaugeProgressBar` from loyalty, rework the component with Circle base svg component, so that we can animate the progress. Also, create a hook that will return animated progress and current color. We only want to animate one animated value, and do all the calculations, based on one reference. Either create more hooks, or extend this one.

sstimac: percentage -> progressPerecentage

components/ProgressRings/ProgressRing.js
```diff
@@ -0,0 +1,85 @@
...
+    halfSize,
+    halfSize,
+    radius,
+    0,
+    currentFillAngle,
+  );
+
+  return (
+    <>
+      {hasBackgroundColor && (
```
sstimac, line 50: check if backgroundWidth is set and > 0

components/ProgressRings/const.js
```diff
@@ -0,0 +1,20 @@
...
+  '#FF4500', // Orange-Red
+  '#FFA500', // Orange
+  '#FFD700', // Yellow
+  '#ADFF2F', // Yellow-Green
+  '#90EE90', // Light Green
+];
+
+export const PROGRESS_RING_DEFAULT_PROPS = {
+  width: 10,
+  backgroundWidth: 10,
```
sstimac, line 12: backgroundLineWidth. 
progressPercentage

components/ProgressRings/const.js
```diff
@@ -0,0 +1,20 @@
...
+  '#FF0000', // Red
+  '#FF4500', // Orange-Red
+  '#FFA500', // Orange
+  '#FFD700', // Yellow
+  '#ADFF2F', // Yellow-Green
+  '#90EE90', // Light Green
+];
+
+export const PROGRESS_RING_DEFAULT_PROPS = {
+  width: 10,
```
sstimac, line 11: progressLineWidth

components/ProgressRings/ProgressRings.js
```diff
@@ -0,0 +1,68 @@
...
+ * percentage values.
+ * It is possible to either specify definite color for each ring by passing ring.color prop,
+ * or give and array of ring.progressColors and component will then interpolate between given
+ * colors, based on given percentage value.
+ */
+const ProgressRings = ({ rings, size, rotation, children, style }) => {
+  const renderRings = useMemo(
+    () =>
+      rings.map((ring, index) => {
+        const ResolvedRingComponent = ring.progressColors
```
sstimac, line 21: Every ring will be animated now

sstimac (changes requested):

components/ProgressRings/ProgressRings.js
tomislav-arambasic: Nope, still using both because either can be used

tomislav-arambasic: ProgressRing can be used by passing definite color, to avoid animations.
AnimatedProgressRing is automatically used when user defines array of `progressColors` parameters for ring and then it's animated - value and color are interpolated.

sstimac (approved):

PR #877, opened 2025-02-12
Feature/add haptic feedback

tomislav-arambasic:
Adds haptic feedback native feature (slight vibration). Excluded in web.


sstimac (approved):
we could go for a minor version bump here, since it invalidates the current native setup

PR #880, opened 2025-02-14
Feature/tab menu component scroll improvements

tomislav-arambasic:
- Add scrolling selected item into view, on mount
- Fix scrolling into view calculation

We could've improved component more by making it functional and by replacing ScrollView with ListView, but my schedule can't afford that now...

https://github.com/user-attachments/assets/484865f3-73b5-416b-ad13-5493386904ac


sstimac (approved):
yeba

PR #935, opened 2026-09-15
feature/ youtube/vimeo video embeds 

s-brankovic:
## Summary
The Video component rendered YouTube and Vimeo sources inside a plain WebView,
which loads the embed without a valid origin (no Referer) and fails with YouTube
"Error 153" — the video area stays blank. This routes YouTube and Vimeo through
their dedicated iframe players so the embed loads correctly. Any other source
(direct file / HTML5) is unchanged.

Fix (patch). Also folds in two small enhancements: YouTube Shorts URLs and an
optional `playerParams` prop.

## Changes
- Video: YouTube → react-native-youtube-iframe, Vimeo → react-native-vimeo-iframe;
  every other source stays on the existing WebView / HTML5 path.
- VideoSourceReader: YouTube URL parsing now also recognizes `shorts/` links;
  added getYouTubeId() / getVimeoId() helpers.
- The iframe players require a numeric height (they can't take '100%'), so height
  is resolved from the `height` prop → the themed container height → a default,
  preserving the previous sizing.
- New optional `playerParams` prop, forwarded to the YouTube player.
- Version 9.2.0 → 9.2.1.


components/Video/VideoSourceReader.js
```diff
@@ -65,4 +65,12 @@ export default class VideoSourceReader {
...
     return this.source;
   }
+
+  getYouTubeId() {
+    return getYouTubeVideoId(this.source);
+  }
+
+  getVimeoId() {
+    return getVimeoVideoId(this.source);
+  }
```
tomislav-arambasic, line 75: I don't see this used anywhere. What's the point of it?

s-brankovic: They are used in components/Video/Video.js in this same PR.

The new YouTube/Vimeo players (react-native-youtube-iframe / react-native-vimeo-iframe) take a video ID rather than an embed URL, so the reader now exposes the ID with getUrl() (which the generic WebView path still uses). Keeping the ID extraction inside VideoSourceReader keeps all the URL-parsing regexes in one place instead of duplicating them in Video.js.

